PR #7, opened 2019-08-05
Add support for Gotify notifications

lukapeschke:
This is a follow-up to https://github.com/containrrr/watchtower/pull/346 , where @simskij proposed to add support gotify notifications to shoutrrr. 

I mostly used the watchtower PR and the microsoft teams service for implementation reference, but I'm not really sure about shoutrrr's implementation patterns, so a review would be nice :smile:

### DIFF
--- a/pkg/services/gotify/gotify.go
+++ b/pkg/services/gotify/gotify.go
@@ -1,0 +1,111 @@
+package gotify
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/containrrr/shoutrrr/pkg/services/standard"
+	"github.com/containrrr/shoutrrr/pkg/types"
+)
+
+// Service providing Gotify as a notification service
+type Service struct {
+	standard.Standard
+	config *Config
+}
+
+// Initialize loads ServiceConfig from configURL and sets logger for this Service
+func (service *Service) Initialize(configURL *url.URL, logger *log.Logger) error {
+	service.Logger.SetLogger(logger)
+	service.config = &Config{}
+	err := service.config.SetURL(configURL)
+	return err
+}
+
+const tokenChars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-_"
+
+// The validation rules have been taken directly from the Gotify source code.
+// These will have to be adapted in case of a change:
+// https://github.com/gotify/server/blob/ad157a138b4985086c484a7aabfc2deada5a33dd/auth/token.go#L8
+func isTokenValid(token string) bool {
+	if len(token) != 15 {
+		return false
+	} else if token[0] != 'A' {
+		return false
+	}
+	for _, c := range token {
+		if !strings.ContainsRune(tokenChars, c) {
+			return false
+		}
+	}
+	return true
+}
+
+func buildURL(config *Config) (string, error) {
+	token := config.Token
+	if len(token) > 0 && token[0] == '/' {
+		token = token[1:]
+	}
+	if !isTokenValid(token) {
+		return "", fmt.Errorf("invalid gotify token \"%s\"", token)
+	}
+	return fmt.Sprintf("https://%s/message?token=%s", config.Host, token), nil
+}
+
+func getPriority(params map[string]string) int {
+	priorityStr, ok := params["priority"]
+	if !ok {
+		priorityStr = "0"
+	}
+	priority, err := strconv.Atoi(priorityStr)
+	if err != nil {
+		priority = 0
+	}
+	return priority
+}
+
+func getTitle(params map[string]string) string {
+	title, ok := params["title"]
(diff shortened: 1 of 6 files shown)


pkg/services/gotify/gotify_test.go
```diff
@@ -0,0 +1,26 @@
...
+	. "github.com/onsi/gomega"
+)
+
+func TestGotify(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Shoutrrr Gotify Suite")
+}
+
+var _ = Describe("the Gotify plugin", func() {
+	It("should work", func() {
```
simskij: Maybe this could be a little more descriptive? What we're actually testing is that the url builder works.

pkg/services/gotify/gotify.go
```diff
@@ -0,0 +1,95 @@
...
+func getTitle(params map[string]string) string {
+	title, ok := params["title"]
+	if !ok {
+		title = "Shoutrrr notification"
+	}
+	return title
+}
+
+// Send a notification message to Microsoft Teams
+func (service *Service) Send(message string, params *types.Params) error {
```
simskij: Would it make more sense from a navigation point if this was placed at the top with the other struct function pointer?

pkg/services/gotify/gotify.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Initialize loads ServiceConfig from configURL and sets logger for this Service
+func (service *Service) Initialize(configURL *url.URL, logger *log.Logger) error {
+	service.Logger.SetLogger(logger)
+	service.config = &Config{}
+	err := service.config.SetURL(configURL)
+	return err
+}
+
+func buildURL(config *Config) (string, error) {
```
simskij: Are there any other format requirements of the token? For example that it should be this and that length or only contain alphanumericals or whatever. If so, that would be nice to validate as well.

pkg/services/gotify/gotify.go
```diff
@@ -0,0 +1,95 @@
...
+	token := config.Token
+	if len(token) < 1 {
+		msg := fmt.Sprintf("invalid gotify token \"%s\"", token)
```
piksel:
```suggestion
		return "", fmt.Errorf("invalid gotify token \"%s\"", token)
```

pkg/services/gotify/gotify.go
```diff
@@ -0,0 +1,95 @@
...
+		return "", errors.New(msg)
```
piksel:
```suggestion
```

pkg/services/gotify/gotify.go
```diff
@@ -0,0 +1,95 @@
...
+
+func getTitle(params map[string]string) string {
+	title, ok := params["title"]
+	if !ok {
+		title = "Shoutrrr notification"
+	}
+	return title
+}
+
+// Send a notification message to Microsoft Teams
```
piksel: Copy-paste error? Should probably refer to this service :)

pkg/services/gotify/gotify_config.go
```diff
@@ -0,0 +1,37 @@
...
+	Prority int
+}
+
+// GetURL returns a URL representation of it's current field values
+func (config *Config) GetURL() *url.URL {
+	return &url.URL{
+		Host:       config.Host,
+		Scheme:     Scheme,
+		ForceQuery: false,
+	}
```
piksel:
```suggestion
		Path:	config.Token,
	}
```

pkg/services/gotify/gotify_test.go
lukapeschke: Completely agree. This was directly "stolen" from the teams part. I also added a few more tests.

pkg/services/gotify/gotify.go
lukapeschke: I like to have my top-level function below the helper functions they call (C habit :) ) I'd prefer to keep it like this if you don't mind

pkg/services/gotify/gotify.go
lukapeschke: Added some checks, based on what I've found in the Gotify code

lukapeschke:
Thanks for the reviews guys! Rebased the branch, added a few unit tests and did the changes you requested. I'm not familiar at all with gingko/gomega though, so please let me know if the test layout could be improved :)

pkg/services/gotify/gotify.go
simskij: Sure. No worries!

pkg/services/gotify/gotify_test.go
simskij: Awesome! Thanks!

simskij:
LGTM! Let's wait for @piksel before merging.
Thanks a lot for the awesome work!

lukapeschke:
Thanks @simskij ! Last edit were just two comments in which I left "teams" instead of "Gotify"

simskij:
Again: thank you for your awesome contribution 👍

simskij:
@all-contributors add @lukapeschke for code and docs

PR #10, opened 2020-01-08
Add support for Pushbullet notifications

MrLuje:


pkg/services/pushbullet/pushbullet_test.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+
+			serviceURL, _ := url.Parse(envPushbulletURL.String())
+			service.Initialize(serviceURL, util.TestLogger())
+			err := service.Send("This is an integration test message", nil)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	Describe("the slack config", func() {
```
simskij: slack -> pushbullet

pkg/services/pushbullet/pushbullet.go
```diff
@@ -69,7 +79,7 @@ func getTargetType(target string) (TargetType, error) {
 
 	if matchesEmail && err == nil {
 		return EmailTarget, nil
-	} else if string(target[0]) == "#" {
+	} else if len(target) > 0 && string(target[0]) == "#" {
```
simskij: else if and else not needed here as all cases return.

simskij:
@all-contributors add @MrLuje for code and docs

PR #32, opened 2020-04-11
switch cli to use cobra and viper

### LINKED ISSUE #12: Feature request: use an environment variable for the URL
It'd be nice to be able to use an environment variable like `SHOUTRRR_URL` or something similar instead of typing it.

I'm suggesting this because command line arguments are logged by Bash in the history file, visible in the process list, etc.

Had a quick look at the source, looks like it could be done without breaking backwards compatibility (assuming that's important now) if we detect the first argument isn't a valid URL so we can interpret it a s the message instead?

simskij:
* Replaces the built-in flags implementation with cobra and viper.
* Also adds environment variable support and cuts away some no longer needed code.
* Adds some rudimentary docs for the cli options and commands.

While this is still very much work in progress, I'd love your feedback on it.

resolves #12


cmd/generate/generate.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Cmd used to generate and display a config from a notification service URL
+var Cmd = &cobra.Command{
+	Use:   "generate",
+	Short: "Generates and displays a config from a notification service URL.",
+	PreRun: func(cmd *cobra.Command, args []string) {
+		// WORKAROUND: make cobra count env vars when checking required flags
+		if hasURLInEnvButNotFlag(cmd) {
+			cmd.Flags().Set("url", viper.GetViper().GetString("url"))
```
piksel, line 27: Yeah, this seems like it should be supported in the libs :/

simskij:
I need to go through this again as I don't think it was ready for merge yet.

piksel:
Yeah, was just wondering what happened to it. And also a bit annoyed that you cannot see merge conflicts on Github :/

simskij:
@piksel please take this for a spin before I merge it.

PR #39, opened 2020-05-30
feat(services): add zulip chat service

arnested:
In our company we just switched to [Zulip Chat](https://zulipchat.com).

What a wonderful opportunity to add another service to Shoutrrr.

### DIFF
--- a/pkg/services/zulip/zulip.go
+++ b/pkg/services/zulip/zulip.go
@@ -1,0 +1,90 @@
+package zulip
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/containrrr/shoutrrr/pkg/services/standard"
+	"github.com/containrrr/shoutrrr/pkg/types"
+)
+
+// Service sends notifications to a pre-configured channel or user
+type Service struct {
+	standard.Standard
+	config *Config
+}
+
+const (
+	contentMaxSize = 10000 // bytes
+	topicMaxLength = 60    // characters
+)
+
+// Send a notification message to Zulip
+func (service *Service) Send(message string, params *types.Params) error {
+	// Clone the config because we might modify stream and/or
+	// topic with values from the parameters and they should only
+	// change this Send().
+	config := service.config.Clone()
+
+	if params != nil {
+		if stream, found := (*params)["stream"]; found {
+			config.Stream = stream
+		}
+
+		if topic, found := (*params)["topic"]; found {
+			config.Topic = topic
+		}
+	}
+
+	topicLength := len([]rune(config.Topic))
+
+	if topicLength > topicMaxLength {
+		return fmt.Errorf(string(TopicTooLong), topicMaxLength, topicLength)
+	}
+
+	messageSize := len(message)
+
+	if messageSize > contentMaxSize {
+		return fmt.Errorf("message exceeds max size (%d bytes): was %d bytes", contentMaxSize, messageSize)
+	}
+
+	return service.doSend(config, message)
+}
+
+// Initialize loads ServiceConfig from configURL and sets logger for this Service
+func (service *Service) Initialize(configURL *url.URL, logger *log.Logger) error {
+	service.Logger.SetLogger(logger)
+	service.config = &Config{}
+
+	if err := service.config.SetURL(configURL); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (service *Service) doSend(config *Config, message string) error {
+	apiURL := service.getURL(config)
+	payload := CreatePayload(config, message)
+	res, err := http.Post(apiURL, "application/x-www-form-urlencoded", strings.NewReader(payload.Encode()))
+
+	if res.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to send notification to service, response status code %s", res.Status)
+	}
+
+	return err
+}
+
+func (service *Service) getURL(config *Config) string {
+	url := url.URL{
+		User:   url.UserPassword(config.BotMail, config.BotKey),
+		Host:   config.Host,
+		Path:   config.Path,
+		Scheme: "https",
+	}
+
+	return url.String()
+}
--- a/pkg/services/zulip/zulip_config.go
+++ b/pkg/services/zulip/zulip_config.go
@@ -1,0 +1,91 @@
+package zulip
+
+import (
+	"errors"
+	"net/url"
+)
+
+// Config for the zulip service
+type Config struct {
+	BotMail string
+	BotKey  string
+	Host    string
+	Path    string
+	Stream  string
+	Topic   string
+}
+
+// GetURL returns a URL representation of it's current field values
+func (config *Config) GetURL() *url.URL {
+	query := &url.Values{}
+
+	if config.Stream != "" {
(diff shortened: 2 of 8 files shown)


pkg/services/zulip/zulip.go
```diff
@@ -0,0 +1,87 @@
...
+	topicMaxLength = 60    // characters
+)
+
+// Send a notification message to Zulip
+func (service *Service) Send(message string, params *types.Params) error {
+	config := *service.config
+
+	if params != nil {
+		if stream, found := (*params)["stream"]; found {
+			config.Stream = stream
```
piksel, line 31: Aren't you mutating the config here? 

That would mean that if you run this two times, the first with params and the second one without, both would use the first ones values.

pkg/services/zulip/zulip_config.go
```diff
@@ -0,0 +1,68 @@
...
+	Topic   string
+}
+
+// GetURL returns a URL representation of it's current field values
+func (config *Config) GetURL() *url.URL {
+	return &url.URL{
+		User:   url.UserPassword(config.BotMail, config.BotKey),
+		Host:   config.Host,
+		Path:   config.Path,
+		Scheme: Scheme,
```
piksel, line 24: The Query parameters are never set (`Stream` and `Topic`)

pkg/services/zulip/zulip_test.go
```diff
@@ -0,0 +1,118 @@
...
+				Expect(serviceConfig).To(Equal(config))
+			})
+		})
+		When("given a config object", func() {
+			It("should build the correct service url", func() {
+				config := zulip.Config{
+					BotMail: "bot-name@zulipchat.com",
+					BotKey:  "correcthorsebatterystable",
+					Host:    "example.zulipchat.com",
+				}
```
piksel, line 106: This test suspiciously does not provide `Stream` and `Topic`  😄

pkg/services/zulip/zulip.go
arnested: I tried to avoid that by dereferencing the pointer to the config in https://github.com/containrrr/shoutrrr/pull/39/files#diff-b282a090227cdf789619987d9821e43fR27

pkg/services/zulip/zulip_config.go
arnested: You are right. Fixed. Thank you.

pkg/services/zulip/zulip_test.go
arnested: You are right. Fixed. Thank you. 😄

pkg/services/zulip/zulip.go
piksel: Well, you are dereferencing the pointer, which means that if `config` changes, no side effects leave the method. But changing `config.*` is not the same thing. 
The problem is that no copy of the structure has taken place, which means that changing the struct fields changes the shared memory.
I might be wrong, and/or juggling too many languages, though 😁, perhaps a test for it is in place regardless?

arnested: I might be wrong, and/or juggling too many languages as well :smile:

I tried it and it worked as I expected.

But since this is not obvious (and we both get confused by switching between languages) I have made an explicit Clone() method on the Config struct and added tests to make sure it's actually a clone and not the same struct:
```patch
diff --git a/pkg/services/zulip/zulip.go b/pkg/services/zulip/zulip.go
index f4e016a..6366138 100644
--- a/pkg/services/zulip/zulip.go
+++ b/pkg/services/zulip/zulip.go
@@ -24,7 +24,10 @@ const (
 
 // Send a notification message to Zulip
 func (service *Service) Send(message string, params *types.Params) error {
-	config := *service.config
+	// Clone the config because we might modify stream and/or
+	// topic with values from the parameters and they should only
+	// change this Send().
+	config := service.config.Clone()
 
 	if params != nil {
 		if stream, found := (*params)["stream"]; found {
@@ -48,7 +51,7 @@ func (service *Service) Send(message string, params *types.Params) error {
 		return fmt.Errorf("message exceeds max size (%d bytes): was %d bytes", contentMaxSize, messageSize)
 	}
 
-	return service.doSend(&config, message)
+	return service.doSend(config, message)
 }
 
 // Initialize loads ServiceConfig from configURL and sets logger for this Service
diff --git a/pkg/services/zulip/zulip_config.go b/pkg/services/zulip/zulip_config.go
index d138343..ff9cb97 100644
--- a/pkg/services/zulip/zulip_config.go
+++ b/pkg/services/zulip/zulip_config.go
@@ -65,6 +65,18 @@ func (config *Config) SetURL(serviceURL *url.URL) error {
 	return nil
 }
 
+// Clone the config to a new Config struct
+func (config *Config) Clone() *Config {
+	return &Config{
+		BotMail: config.BotMail,
+		BotKey:  config.BotKey,
+		Host:    config.Host,
+		Path:    config.Path,
+		Stream:  config.Stream,
+		Topic:   config.Topic,
+	}
+}
+
 const (
 	// Scheme is the identifying part of this service's configuration URL
 	Scheme = "zulip"
diff --git a/pkg/services/zulip/zulip_test.go b/pkg/services/zulip/zulip_test.go
index 22939a3..65f6762 100644
--- a/pkg/services/zulip/zulip_test.go
+++ b/pkg/services/zulip/zulip_test.go
@@ -79,6 +79,36 @@ var _ = Describe("the zulip service", func() {
 		})
 	})
 	Describe("the zulip config", func() {
+		When("cloning a config object", func() {
+			It("the clone should have equal values", func() {
+				config1 := &zulip.Config{
+					BotMail: "bot-name@zulipchat.com",
+					BotKey:  "correcthorsebatterystable",
+					Host:    "example.zulipchat.com",
+					Path:    "api/v1/messages",
+					Stream:  "foo",
+					Topic:   "bar",
+				}
+
+				config2 := config1.Clone()
+
+				Expect(config1).To(Equal(config2))
+			})
+			It("the clone should not be the same struct", func() {
+				config1 := &zulip.Config{
+					BotMail: "bot-name@zulipchat.com",
+					BotKey:  "correcthorsebatterystable",
+					Host:    "example.zulipchat.com",
+					Path:    "api/v1/messages",
+					Stream:  "foo",
+					Topic:   "bar",
+				}
+
+				config2 := config1.Clone()
+
+				Expect(config1).NotTo(BeIdenticalTo(config2))
+			})
+		})
 		When("generating a config object", func() {
 			It("should generate a correct config object", func() {
 				zulipURL, err := url.Parse("zulip://bot-name%40zulipchat.com:correcthorsebatterystable@example.zulipchat.com?stream=foo&topic=bar")
```

simskij:
Then I think we're done here! Good job! Merging.

pkg/services/zulip/zulip.go
piksel: Yeah, I was wrong. When doing a short variable declaration of a struct, you are actually making a copy of that struct. Which does make sense, since it's fields are actually part of the variable, and not pointers. So you would only need to do the manual cloning when the struct itself contains references, like slices for example.

arnested: Well, it _is_ confusing. I think being more explicit as it ended up being is way better so thanks for wondering :+1:

PR #44, opened 2020-06-07
feat(smtp): add support for oauth2 and gmail generators

piksel:
The generator entry points are temporary until #32 is done.

A better API for service generators should be probably be added.


piksel:
I don't think there is much more to add here. This supports XOAUTH2 authentication for the consumers that want it. More built in providers could be added, but it's not something I am doing anytime soon (unless there is a explicit need for it).

PR #60, opened 2020-08-27
Add proxy documentation

piksel:


piksel:
How silly of me, I should of course TEST the documentation as well.

PR #61, opened 2020-09-03
fix(services/teams): parsing 'teams://' URLs incorrectly

andoks:
Currently the teams service is parsing the teams:// URLs incorrectly (ref [docs](https://github.com/containrrr/shoutrrr/blob/master/docs/services/teams.md)).

- token A has the shape `<UUID>@<UUID>`
- token B has the shape `<UUID>`
- token C has the shape `<UUID>`

Translated to the teams:// URL scheme, that results in `teams://<UUID>@<UUID>/UUID/UUID`

This means that:
- token A is the concatenation of userinfo and hostname with an '@' in between when parsed using url.URL.
- token B is the dir with the slash removed
- token C is the "file" when treating url.URL.Path as a file path

### DIFF
--- a/pkg/services/teams/teams_config.go
+++ b/pkg/services/teams/teams_config.go
@@ -1,7 +1,9 @@
 package teams
 
 import (
+	"fmt"
 	"net/url"
+	"path"
 
 	"github.com/containrrr/shoutrrr/pkg/services/standard"
 )
@@ -25,10 +27,9 @@ func (config *Config) GetURL() *url.URL {
 
 // SetURL updates a ServiceConfig from a URL representation of it's field values
 func (config *Config) SetURL(url *url.URL) error {
-
-	tokenA := url.User.Username()
-	tokenB, _ := url.User.Password()
-	tokenC := url.Hostname()
+	tokenA := fmt.Sprintf("%v@%v", url.User.Username(), url.Hostname())
+	_, tokenB := path.Split(path.Dir(url.Path))
+	tokenC := path.Base(url.Path)
 
 	config.Token = Token{
 		A: tokenA,


andoks:
When trying to send notifications without this change, shoutrrr errors out with this error: `error invoking send: failed to send notification to teams, response status code 400 Bad Request`

andoks:
[teams_test.go]() uses API URL format that is different from the one I received when creating webhook (teams://88888888-4444-333-333-cccccccccccc:11111111112222222222333333333344@88888888-4444-333-333-cccccccccccc resulting in API call https://outlook.office.com/webhook/88888888-4444-333-333-cccccccccccc/IncomingWebhook/11111111112222222222333333333344/88888888-4444-333-333-cccccccccccc)

andoks:
This PR is the minimal change needed to make shoutrrr work for the URL I got from teams.

I have also drafted a change to the teams service that should be able to handle any URLs by doing less parsing in the teams service of the URL provided, with the only modification being the scheme being `teams+https` instead of plain `https` to actually be able to decide to use the teams service for handling those URLs when sending messages ([branch here](https://github.com/andoks/shoutrrr/tree/topic/fix_teams_service_url_parsing_rewrite_url_passing))

piksel:
Yeah, this is really confusing. But the URL in the test is for testing an inacurrate URL. It's really lacking though.
The documented URL seems to line up with your results, so I'm not sure what is going on. I tried the teams service pretty recently, and it worked correctly then. I will investigate this further.

simskij:
@piksel, any thoughts on how to proceed with this one?

piksel:
My guess is that the documentation was never updated when the teams-URL was changed. To avoid breaking consumers configuration I would avoid changing the URL. even though the current one is pretty bad (and is not documented properly).
This could be changed so that it detects both versions and perhaps emits a warning about the current one.

The best method is probably to add support for the `teams+https` one as that one doesn't unnecessarily complicate the URL. It could be added as another service in the router to support both.

@andoks is that a good solution for you?

andoks:
> My guess is that the documentation was never updated when the teams-URL was changed. To avoid breaking consumers configuration I would avoid changing the URL. even though the current one is pretty bad (and is not documented properly).
> This could be changed so that it detects both versions and perhaps emits a warning about the current one.
> 
> The best method is probably to add support for the `teams+https` one as that one doesn't unnecessarily complicate the URL. It could be added as another service in the router to support both.
> 
> @andoks is that a good solution for you?

That would be a good solution for me, but I have put the feature I was thinking of using this for on hold for now, meaning I'm not planning on using shoutrrr in the forseable future.

piksel:
This is pretty much superseded by #117, so closing it in favor of that.

Issue #45, opened 2020-06-07
Issue with telegram format: first path segment in URL cannot contain colon\n


oscargb1:
I'm using the Shoutrrr integration in Watchtower to send notifications via Telegram, but it seems the format of the notification URL that I'm using isn't correct, as it's giving me the following issue:

`time="2020-06-07T02:33:26+02:00" level=fatal msg="Failed to initialize Shoutrrr notifications: error initializing router services: parse \"\\\"telegram://<bot_token>@telegram?channels=<chat_id>\\\"\": first path segment in URL cannot contain colon\n"`

As there isn't a detailed documentation for the Telegram notification, can you tell me if I'm using the correct format?

piksel:
I think you need a slash between `@telegram` and `?channel`:
```
telegram://<bot_token>@telegram/?channels=<chat_id>
```

oscargb1:
Yeah, I noticed that on the telegram tests. But it's still giving me the issue with that change.

simskij:
Could you please paste the full docker run command you're using, but replace the right hand of your token with `mock-token`?

I'm unable to reproduce your error, even without putting a `/` before the `?`

oscargb1:
I'm using docker compose:
  watchtower:
    container_name: watchtower
    image: containrrr/watchtower
    restart: always
    environment:
     - WATCHTOWER_NOTIFICATIONS=shoutrrr
     - WATCHTOWER_NOTIFICATION_URL="telegram://632861895:mock-token@telegram/?channels=618771099"
     - TZ=Europe/Madrid
    volumes:
     - /var/run/docker.sock:/var/run/docker.sock
    command: --debug --interval 3600 --cleanup

simskij:
Ah! You need to change your yaml 😅 

Either:
```yaml
- WATCHTOWER_NOTIFICATION_URL=telegram://632861895:mock-token@telegram/?channels=618771099
```

or
```yaml
  WATCHTOWER_NOTIFICATION_UR: "telegram://632861895:mock-token@telegram/?channels=618771099"
```

oscargb1:
Oops 🙈. Sorry I'm new to Docker.
Thank you so much for the help, now it works!

PR #62, opened 2020-09-04
(feat): auto reads proxy from environment.

leopku:
When HTTP_PROXY and HTTPS_PROXY was set, shoutrrr should auto use it. 
Otherwise, shoutrrr should act as normal.

Related: #45

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -3,6 +3,7 @@ package router
 import (
 	"fmt"
 	"log"
+	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


piksel:
`ProxyFromEnvironment` is used in the [default client](https://github.com/golang/go/blob/go1.15/src/net/http/transport.go#L43) already. This will just override the proxy settings if the consumer has already overriden the proxy settings.

leopku:
Sorry for not notice that it is already there.
Thanks:)

PR #67, opened 2020-10-21
Correct the ID part of webhook

alii:
The ID is not actually the channel, it's the ID of the webhook – all webhooks are technically bot users under the hood, and this is their user ID (snowflake)


simskij (approved):
Thank you for your contribution! 🎉

PR #69, opened 2020-11-01
feat(smtp): add support for implicit TLS

### LINKED ISSUE #68: [Feature request] Add support for implicit TLS (port 465) for SMTP
I tried using shoutrrr with an SMTP URL that's using implicit TLS (port 465) earlier today and observed that the request timed out after a while. The same URL worked fine when using explicit TLS on port 587 via STARTTLS.

piksel:
Add SMTPS (implicit TLS) support for SMTP.
Use implicit TLS by default if port is set to `465`, configurable by setting the `Encryption` field.

Fixes #68

### DIFF
--- a/pkg/services/smtp/smtp.go
+++ b/pkg/services/smtp/smtp.go
@@ -38,6 +38,7 @@ func (service *Service) Initialize(configURL *url.URL, logger *log.Logger) error
 		Auth:        authTypes.Unknown,
 		UseStartTLS: true,
 		UseHTML:     false,
+		Encryption:  encMethods.Auto,
 	}
 	if err := service.config.SetURL(configURL); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


pkg/services/smtp/smtp.go
```diff
@@ -59,15 +60,26 @@ func (service *Service) Send(message string, params *types.Params) error {
 	if params == nil {
 		params = &types.Params{}
 	}
-	client, err := getClientConnection(service.config.Host, service.config.Port)
+	client, err := getClientConnection(service.config.Host, service.config.Port, service.config.Encyption)
```
simskij, line 63:
```suggestion
	client, err := getClientConnection(service.config)
```

This would require some updates of the function as well, but I think you get the gist. 😅 In the spirit of maintainability, we should refrain from implementing triadic functions if possible.

pkg/services/smtp/smtp.go
```diff
@@ -59,15 +60,26 @@ func (service *Service) Send(message string, params *types.Params) error {
...
-	conn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", host, port))
+func getClientConnection(host string, port uint16, encryption encMethod) (*smtp.Client, error) {
+
+	var conn net.Conn
+	var err error
+
+	if useImplicitTLS(encryption, port) {
+		conn, err = tls.Dial("tcp", fmt.Sprintf("%s:%d", host, port), &tls.Config{
+			ServerName: host,
+		})
```
simskij, line 78:
```suggestion
        url := fmt.Sprintf("%s:%d", host, port)
        tlsConfig := &tls.Config{ServerName: host}
        
		conn, err = tls.Dial("tcp", url, tlsConfig)
```

I know you use GoLand, but I assume most readers don't. 😄

simskij (changes requested):
LGTM. Some minor nitpicks

pkg/services/smtp/smtp.go
piksel: It's for testing purposes I think, but I'll look into it.

pkg/services/smtp/smtp.go
piksel: split out the Sprintf to `addr` variable. the tls-config is verbose enough IMO.

pkg/services/smtp/smtp.go
piksel: Nope. Changed to just requiring `config`.

PR #73, opened 2020-11-10
feat: add support for opsgenie

JonasPf:


pkg/services/opsgenie/opsgenie_json.go
```diff
@@ -0,0 +1,75 @@
...
+// way we can pass in complex types as JSON like so:
+//
+//	service.Send("An example alert message", &types.Params{
+//		"alias":       "Life is too short for no alias",
+//		"description": "Every alert needs a description",
+//		"responders":  `[{"id":"4513b7ea-3b91-438f-b7e4-e3e54af9147c","type":"team"},{"name":"NOC","type":"team"}]`,
+//		"visibleTo":   `[{"id":"4513b7ea-3b91-438f-b7e4-e3e54af9147c","type":"team"},{"name":"rocket_team","type":"team"}]`,
+//		"details":     `{"key1": "value1", "key2": "value2"}`,
+//	})
+type JSON struct {
```
piksel, line 24: Using the arbitrary type name `JSON` for this struct is a bit confusing. Perhaps something like `AlertPayload` or `AlertBody`?

pkg/services/opsgenie/opsgenie_test.go
```diff
@@ -0,0 +1,134 @@
...
+const (
+	mockAPIKey = "eb243592-faa2-4ba2-a551q-1afdf565c889"
+)
+
+func TestOpsGenie(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Shoutrrr OpsGenie Suite")
+}
+
+var _ = Describe("the OpsGenie plugin", func() {
```
piksel, line 29:
```suggestion
var _ = Describe("the OpsGenie service", func() {
```
Nitpick: we try to consistently use the phrase service

pkg/services/opsgenie/opsgenie_config.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/url"
+
+	"github.com/containrrr/shoutrrr/pkg/services/standard"
+)
+
+// Config for use within the opsgenie plugin
+type Config struct {
+	ApiKey string `desc:"The OpsGenie API key"`
+	Host   string `desc:"The OpsGenie API host. Use 'api.opsgenie.com' for US and 'api.eu.opsgenie.com' for EU instances"`
+	standard.QuerylessConfig
```
piksel, line 13: The params on send should probably (at least some of them) be settable using the query string.

piksel (commented):
Looks good, but I have added some feedback.

JonasPf:
Thanks for the feedback. I won't have time within the next 2 weeks but I'll get back to it after that to fix those things.

JonasPf:
Hi All,

I pushed changes that address the issues mentioned above. Apparently I also rebased wrong or something like that because I see a few commits that probably shouldn't be in this branch. What's your general policy here? I can fix this if you are ok with a force-push to this branch. Alternatively I could open a new PR.

piksel:
I would squash this anyway, so it should be fine!

pkg/services/opsgenie/opsgenie_config.go
```diff
@@ -0,0 +1,98 @@
...
+
+	for key, vals := range url.Query() {
+		if err := config.Set(key, vals[0]); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
```
piksel, line 94: This doesn't implement `GetURL` which would mean that it cannot be used with `shoutrrr generate`. It's a small addition though:
```suggestion

func (config *Config) GetURL() *url.URL {
	return &url.URL{
		Host:       fmt.Sprintf("%s:%d", config.Host, config.Port),
		Path:       fmt.Sprintf("/%s", config.ApiKey),
		Scheme:     Scheme,
		ForceQuery: true,
		RawQuery:   format.BuildQuery(config),
	}
}

```

docs/services/opsgenie.md
```diff
@@ -0,0 +1,60 @@
...
+    "priority":    "P1",
+    "user":        "Dracula",
+    "note":        "Here is a note",
+})
+```
+
+# Optional parameters
+
+You can optionally specify the parameters in the URL:
+pushover://shoutrrr:token@userKey/?devices=device&title=Custom+Title&priority=1
```
piksel, line 54: This seems.... wrong 😁
```suggestion
opsgenie://api.opsgenie.com/eb243592-faa2-4ba2-a551q-1afdf565c889?alias=Life+is+too+short+for+no+alias&description=Every+alert+needs+a+description&actions=An+action&tags=["tag1","tag2"]&entity=An+example+entity&source=The+source&priority=P1&user=Dracula&note=Here+is+a+note
```

pkg/services/opsgenie/opsgenie_json.go
```diff
@@ -0,0 +1,92 @@
...
+// or maps. Instead we "preserve" the JSON in those fields. That
+// way we can pass in complex types as JSON like so:
+//
+//	service.Send("An example alert message", &types.Params{
+//		"alias":       "Life is too short for no alias",
+//		"description": "Every alert needs a description",
+//		"responders":  `[{"id":"4513b7ea-3b91-438f-b7e4-e3e54af9147c","type":"team"},{"name":"NOC","type":"team"}]`,
+//		"visibleTo":   `[{"id":"4513b7ea-3b91-438f-b7e4-e3e54af9147c","type":"team"},{"name":"rocket_team","type":"team"}]`,
+//		"details":     `{"key1": "value1", "key2": "value2"}`,
+//	})
```
piksel, line 23: Using raw JSON in the Query and params is not ideal. If the user know the API well and types JSON frequently it might work, but there is no feedback that something might not work.

Arrays are more or less universally supplied as `foo,bar,baz` in Shoutrrr services, which is easy to de-/serialize, but the others might need some more special treatment.

Perhaps responders/visibleTo could be supplied as 
`&responders=team:4513b7ea-3b91-438f-b7e4-e3e54af9147c,team:NOC` and `&visibleTo=team:4513b7ea-3b91-438f-b7e4-e3e54af9147c,team:rocket_team`?
And perhaps the details could just be: `details=key1:value1,key2:value2` ?

It's easier to manually input and since it's parsed we can actually give feedback to the user about the format and show an appropriate error when it's not in the correct format.
It will also be work better with `shoutrrr generate` and `shoutrr verify`.

With #92 you can get a lot of the de-/serialization of the params and query strings for free, so I guess this can be fixed then. I probably don't want to make a new release with OpsGenie and the raw JSON syntax though, but I can help with suppling the methods for reading and writing those fields.

piksel (changes requested):
A lot of great work has been put into this, but I'm not so sure about the raw JSON URLs.
Left some comments.

simskij:
Alright, closing this for the time being then. We try to aim at only implementing services that people actually will use. If anyone actually wants/needs Opsgenie support, please open another ticket and let us know.

JonasPf:
> Alright, closing this for the time being then. We try to aim at only implementing services that people actually will use. If anyone actually wants/needs Opsgenie support, please open another ticket and let us know.
> 
> Thanks for your contributions!

Well, I'm using it, that's why I built it in the first place ;-)

I'm happy work work on the remaining issues if you plan on merging it, just let me know. Until then, it works well enough for me as it is.

simskij:
I must have mixed issues together! I closed this because I was under the impression that you abandoned the idea of using shoutrrr for your project. 

Reopening.

docs/services/opsgenie.md
JonasPf: Yeah, it's a bit of a monster query. I just took the examples from the OpsGenie docs and converted them to query params.

IMO there is benefit in showing all supported options. Would it be helpful to split it up into several examples?

pkg/services/opsgenie/opsgenie_json.go
JonasPf: That should work for arrays and maps.

For responders/visibleTo we need to distinguish between a name or an id. Scroll down a bit here for all options: https://docs.opsgenie.com/docs/alert-api#create-alert. It's not clear to me how to do that with your example. 

How about this adding the name/id information to the key like this?

```
&responders=team-id:4513b7ea-3b91-438f-b7e4-e3e54af9147c,team-name:NOC,schedule-id:XXXX
```

JonasPf:
> I must have mixed issues together! I closed this because I was under the impression that you abandoned the idea of using shoutrrr for your project.
> 
> Reopening.

No worries, I know I'm sometimes a bit slow to respond. Especially around Christmas :-)

pkg/services/opsgenie/opsgenie_json.go
JonasPf: Are there other services in OpsGenie that need similar complex parameters?

After thinking about it a bit, I believe that most people won't need to set `responders/visibleTo` as URL parameter. The OpsGenie UI/API offer much more sophisticated ways of routing alerts to teams/users. Compared to setting URL parameters, this lets users configure the routing rules centrally, instead of sending them in via every alert.

Still, no harm to get this to work right but using this is probably special case rather then the norm.

docs/services/opsgenie.md
piksel: Well, I just meant that the original row was saying "pushover" which... isn't this service 😁

pkg/services/opsgenie/opsgenie_json.go
piksel: My example just presumed that all IDs would be in the GUID-ish format, and if it matched that format it would be interpreted as an ID.

If you for some bizarre reason would have a team with the name `4513b7ea-3b91-438f-b7e4-e3e54af9147c`, then you'd have to specify that one by ID instead. That seems like it should cover all cases. But perhaps the IDs can be in another format?

docs/services/opsgenie.md
JonasPf: Oh, ok got it ;-)

pkg/services/opsgenie/opsgenie_json.go
JonasPf: I haven't seen anything other then this format so far in the 6 months that I've used OpsGenie. Having said that, there is always the slight possibility that OpsGenie might change their id format in the future. Unlikely but possible.

If you are ok with that, I can go and implement it using your proposed solution.

JonasPf:
I just pushed some changes. The service now works well with the `generate` and `verify` commands, has a couple more tests and I merged all the changes from master.

Next I'll be working on integrating PropKeyResolver and replacing the raw json fields with more human readable format as suggested by @piksel.

JonasPf:
Took a while but here is another update. It now uses more shoutrrr-like formatting for the config values:

- `actions` and `tags` are simple lists and I'm using the existing serialization logic.
- `details` is an object/map. I figured that that's a common enough data type to add it to `formatter.go`
- `responders` and `visibleTo` are very much OpsGenie specific (and non-trivial to serialize) so I implemented them in `entity.go` in the Opsgenie package.

It mostly works but I'm running into a few problems that make me wonder if I'm on the right track:

1. Commas in the URL can cause issues with the `send` cli command. It is possible to set multiple URLs separated by commas (https://github.com/containrrr/shoutrrr/blob/main/cli/cmd/send/send.go#L42). If an URL already contains a comma it gets split in the wrong place. Because Cobra internally parses its inputs as CSV, it is possible to wrap every URL in quotes to get around this. When using bash, these quotes need to be escaped to prevent bash from interpreting them. Is it meant like this?

2. How would you deal with spaces in the URL? In web URLs they are encoded with '+' or '%20'. Are there any other services that have fields like `description` that need spaces? The code works with normal unencoded spaces at the moment.

3. JSON field values like `priority`, `responders`, etc. can be set either via URL or at runtime by passing them in as parameters. The problem is that automatic (de-)serialization is only supported for a Config struct, not runtime  parameters. As a result I had to duplicated some of the (de-)serialization code in `opsgenie_json.go`. Additionally there is the OpsGenie specific (de-)serialization code in `entity.go`. I'm not super happy about having similar logic in 3 different places. What do you think? Is there a better way to implement this?

4. Related to 3: The `generate` cli command doesn't work with the complex datatypes like `responders` and `visibleTo`. I think the reasons is simply that the `generate` command uses the de-serialization code in `formatter.go` and not the one in `entity.go` - making me again wonder if there is a better way to implement this.

piksel:
> Commas in the URL can cause issues with the send cli command. It is possible to set multiple URLs separated by commas (https://github.com/containrrr/shoutrrr/blob/main/cli/cmd/send/send.go#L42). If an URL already contains a comma it gets split in the wrong place. Because Cobra internally parses its inputs as CSV, it is possible to wrap every URL in quotes to get around this. When using bash, these quotes need to be escaped to prevent bash from interpreting them. Is it meant like this?

Commas are used elsewhere as well, send is going to stop interpreting the commas as multiple arguments (don't remember which branch I fixed that in).
> 
> How would you deal with spaces in the URL? In web URLs they are encoded with '+' or '%20'. Are there any other services that have fields like description that need spaces? The code works with normal unencoded spaces at the moment.
Spaces are generally fine, except for in CLI and watchtower environment variables. But you can use either of the URL escapes for spaces if needed (the URL parser handles that).

> JSON field values like priority, responders, etc. can be set either via URL or at runtime by passing them in as parameters. The problem is that automatic (de-)serialization is only supported for a Config struct, not runtime parameters. As a result I had to duplicated some of the (de-)serialization code in opsgenie_json.go. Additionally there is the OpsGenie specific (de-)serialization code in entity.go. I'm not super happy about having similar logic in 3 different places. What do you think? Is there a better way to implement this?

There is automatic de-/serializing of properties (query parameters and send params) that can be used at runtime, but I'll take a look to see if that can be used.

> Related to 3: The generate cli command doesn't work with the complex datatypes like responders and visibleTo. I think the reasons is simply that the generate command uses the de-serialization code in formatter.go and not the one in entity.go - making me again wonder if there is a better way to implement this.

Yeah, the formatter serialization have no way of handling structs atm,. but it should be pretty easy to add an interface that they can implement to serialize to/from a string.

pkg/services/opsgenie/opsgenie_config.go
```diff
@@ -0,0 +1,121 @@
...
+		visibleTo, _ := serializeEntities(config.VisibleTo)
+		rawQuery = rawQuery + separator + "visibleTo=" + visibleTo
+	}
+
+	result := &url.URL{
+		Host:     host,
+		Path:     fmt.Sprintf("/%s", config.ApiKey),
+		Scheme:   Scheme,
+		RawQuery: rawQuery,
+	}
```
piksel, line 74: This could be made a little bit simpler by using the `url.Values` for manipulating the query:
```suggestion
	// Serialize Shoutrrr standard data types
	query, _ := url.ParseQuery(format.BuildQuery(resolver))

	// Serialize OpsGenie specific data types
	if len(config.Responders) > 0 {
		responders, _ := serializeEntities(config.Responders)
		query.Set("responders", responders)
	}
	if len(config.VisibleTo) > 0 {
		visibleTo, _ := serializeEntities(config.VisibleTo)
		query.Set("visibleTo", visibleTo)
	}

	result := &url.URL{
		Host:     host,
		Path:     fmt.Sprintf("/%s", config.ApiKey),
		Scheme:   Scheme,
		RawQuery: query.Encode(),
	}
```

pkg/services/opsgenie/opsgenie_json.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		if err != nil {
+			return result, err
+		}
+
+	}
+
+	return result, nil
+}
```
piksel, line 129: The propkeyresolver can update the config from params (which means that the config should be cloned before doing this), but it will stop and return an error for the structs, so you can just update them first and remove them from the map:

```suggestion
func NewAlertPayload(message string, config *Config, params *types.Params) (AlertPayload, error) {
	if params == nil {
		params = &types.Params{}
	}

	pkr := format.NewPropKeyResolver(config)

	if value, found := (*params)["responders"]; found {
		responders, _ := deserializeEntities(value)
		config.Responders = responders
		delete(*params, "responders")
	}

	if value, found := (*params)["visibleTo"]; found {
		responders, _ := deserializeEntities(value)
		config.Responders = responders
		delete(*params, "visibleTo")
	}

	if err := pkr.UpdateConfigFromParams(config, params); err != nil {
		return AlertPayload{}, err
	}

	result := AlertPayload{
		Message: message,
		// Populate with values from the query string as defaults
		Alias:       config.Alias,
		Description: config.Description,
		Responders:  config.Responders,
		VisibleTo:   config.VisibleTo,
		Actions:     config.Actions,
		Tags:        config.Tags,
		Details:     config.Details,
		Entity:      config.Entity,
		Source:      config.Source,
		Priority:    config.Priority,
		User:        config.User,
		Note:        config.Note,
	}

	return result, nil
}
```

piksel (approved):
Great work overall! The suggested changes are optional, but will increase the test coverage a bit if nothing else :)
Also; I merged the send `ArraySlice` fix, so now you can correctly test using the CLI (if you rebase/merge `main`)

pkg/services/opsgenie/opsgenie_config.go
JonasPf: I tried that. The problem is that `query.Encode()` encodes commas as well so something like `tags=t1,t2` becomes tags=t1%2ct2`

pkg/services/opsgenie/opsgenie_json.go
JonasPf: Brilliant, that makes everything simpler. I've added it with some minor moidifications. Not sure what you mean by

> Doing the param update should probably be moved to opsgenie.go since Service already has a PropKeyResolver instance ready to use.

I couldn't find a `PropKeyResolver` in `opsgenie.go`. There is a transient one in `opsgenie_config.go` that I could potentially save and re-use.

JonasPf:
I merged the changes in master. Problem is that `formatQuery.BuildQuery()` got changed to use `Values.Encode()`. This breaks a test because it encode a comma. Same problem as mentioned above. I left the failing code in there for you to see it.

---

`generate` doesn't work with the `Responders` and `VisibleTo` fields. It keeps saying `Invalid format for field Responders: field format is not supported` which makes sense because `formatter.go` doesn't know how to deserialize entities.
I don't know how to solve that yet. Need to think about it. Any ideas?

Btw. I found two unrelated issues:

- https://github.com/containrrr/shoutrrr/issues/135
- https://github.com/containrrr/shoutrrr/issues/134

pkg/services/opsgenie/opsgenie_test.go
```diff
@@ -0,0 +1,417 @@
...
+			err = config.SetURL(url)
+			Expect(err).To(BeNil())
+
+			Expect(config.Port).To(Equal(uint16(12345)))
+		})
+	})
+
+	When("generating a config from a url with query parameters", func() {
+		It("should populate the config fields with the query parameter values", func() {
+			queryParams := `alias=Life+is+too+short+for+no+alias&description=Every+alert+needs+a+description&actions=An+action&tags=tag1,tag2&details=key:value,key2:value2&entity=An+example+entity&source=The+source&priority=P1&user=Dracula&note=Here+is+a+note&responders=user:Test,team:NOC&visibleTo=user:A+User`
```
piksel, line 316: This would make the test work with the encoded query field values:
```suggestion
			queryParams := `alias=Life+is+too+short+for+no+alias&description=Every+alert+needs+a+description&actions=An+action&tags=tag1%2Ctag2&details=key%3Avalue%2Ckey2%3Avalue2&entity=An+example+entity&source=The+source&priority=P1&user=Dracula&note=Here+is+a+note&responders=user%3ATest%2Cteam%3ANOC&visibleTo=user%3AA+User`
```
The users won't have to escape all the characters, it will handle both, but the generated ones are at least ensured to always be valid URL query params.

piksel (approved):
Nice work!

I will look into creating an interface for de-/serializing structs that the entities can implement.

If you move out the looping from the entity methods to a pair of:
```go
(e *Entity) SetFromProp(propValue string) error {  }
```
and 
```go
(e *Entity) GetPropValue() (string, error) {  }
```
I will try to get it implemented ASAP

pkg/services/opsgenie/opsgenie_json.go
piksel: No, sorry, usually there is an instance of `PropKeyResolver` stored on the `Service` struct for the services. That's because it's used at least once (often twice) in initializing, for setting the config defaults and for updating the config according to the supplied URL. 
Example:
https://github.com/containrrr/shoutrrr/blob/main/pkg/services/telegram/telegram.go#L30-L56

pkg/format/format_query.go
```diff
@@ -19,7 +19,7 @@ func BuildQuery(cqr types.ConfigQueryResolver) string {
 		}
 		value, err := cqr.Get(key)
 
-		if err == nil {
+		if err == nil && value != "" {
```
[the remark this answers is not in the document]
JonasPf: Fixed the test and added another one to test when values are populated.

pkg/services/opsgenie/opsgenie_config.go
```diff
@@ -0,0 +1,121 @@
...
+	return map[string]types.EnumFormatter{}
+}
+
+// Public version of GetURL that creates a new PropKeyResolver when accessed from outside the package
+func (config *Config) GetURL() *url.URL {
+	resolver := format.NewPropKeyResolver(config)
+	return config.getURL(&resolver)
+}
+
+// Private version of GetURL that can use an instance of PropKeyResolver instead of rebuilding it's model from reflection
```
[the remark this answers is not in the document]
JonasPf: Yeah, that's because I copied it from somewhere else :blush:

pkg/services/opsgenie/opsgenie_test.go
JonasPf: Ok fixed. I misunderstood and thought you wanted to preserve the commas, etc. when generating urls.

pkg/services/opsgenie/opsgenie_config.go
JonasPf: Ok fixed

pkg/services/opsgenie/opsgenie_json.go
JonasPf: Got it, fixed

pkg/services/opsgenie/opsgenie_test.go
JonasPf: > If you move out the looping from the entity methods to a pair of

Done, is that what you had in mind?

piksel: > If you move out the looping from the entity methods to a pair of:
That was one mess of a sentence. But yeah, exactly.

JonasPf:
Hey, I'm still aware of this and want to finally test it, just haven't had the time yet. It's not forgotten!

JonasPf:
Ok , so I merged #137 and tested everything again. `generate` and `verify` now work. One tiny thing I noticed: `formatter` lowercases all keys. What this means is that `visibleTo` becomes `visibleto` after running `generate`. This only seems to affect how the user sees the URL. It still gets send to OpsGenie as `visibleTo`.

I also manually tested sending a bunch of differently configured alerts to OpsGenie. That worked as expected.

Functionality wise I this is ready to be merged from my perspective.

piksel:
Yeah, exactly. This is kind of the same deal as with the commas etc. The generated URLs have the query values in alphabetical order, uses lowercase keys and are escaped. This is mainly for testing purposes, if the format is uniform and known beforehand, black box tests for the serialization is easier to write and update when something changes.

I'll fix the CA issues and merge #137, hopefully this won't have any conflicts after that, in which case I'll merge this as well. Great job!

piksel:
Solved merge conflicts and merged as #140!

Great job! 🎉

Issue #70, opened 2020-11-05
If user formats rawURL to contain '#' before channel name - execution ends up in segfault


atighineanu:
**When I tried** shoutrrr code for a rocket chat automated notification, and formatted the channel name with a '#' prefix - that is what I got:
```
go run cli/main.go send --url "rocketchat://shoutrrr@<myhost>/<tokenA>/<tokenB>/#general" --message "hello hello"
panic: runtime error: slice bounds out of range [:1] with length 0
goroutine 1 [running]:
github.com/containrrr/shoutrrr/pkg/services/rocketchat.(*Config).SetURL(0xc0000c27e0, 0xc0000d8120, 0x1548080, 0xc0000b8980)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/pkg/services/rocketchat/rocketchat_config.go:69 +0x77d
github.com/containrrr/shoutrrr/pkg/services/rocketchat.(*Service).Initialize(0xc0000b8980, 0xc0000d8120, 0xc0000c8280, 0xa, 0xc0000b1938)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/pkg/services/rocketchat/rocketchat.go:24 +0x85
github.com/containrrr/shoutrrr/pkg/router.(*ServiceRouter).initService(0x191ce20, 0x7ffeefbff694, 0x73, 0x15aa11a, 0x6, 0x15d0868, 0x1501b60)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/pkg/router/router.go:171 +0xe6
github.com/containrrr/shoutrrr/pkg/router.(*ServiceRouter).Locate(...)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/pkg/router/router.go:203
github.com/containrrr/shoutrrr.Send(0x7ffeefbff694, 0x73, 0x7ffeefbff712, 0x15, 0x15, 0x0)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/shoutrrr.go:19 +0x45
github.com/containrrr/shoutrrr/cli/cmd/send.Run(0x1914580, 0xc0000dce00, 0x0, 0x4)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/cli/cmd/send/send.go:49 +0x146
github.com/spf13/cobra.(*Command).execute(0x1914580, 0xc0000dcdc0, 0x4, 0x4, 0x1914580, 0xc0000dcdc0)
[...]

PR #74, opened 2020-11-13
fix for https://github.com/containrrr/shoutrrr/issues/70

ellisab:
Signed-off-by: Alexandru Bonini <boninialexandru@gmail.com>

### DIFF
--- a/pkg/services/rocketchat/rocketchat_config.go
+++ b/pkg/services/rocketchat/rocketchat_config.go
@@ -3,9 +3,10 @@ package rocketchat
 import (
 	"errors"
 	"fmt"
-	"github.com/containrrr/shoutrrr/pkg/services/standard"
 	"net/url"
 	"strings"
+
+	"github.com/containrrr/shoutrrr/pkg/services/standard"
 )
 
 // Config for the rocket.chat service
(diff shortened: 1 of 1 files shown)


piksel (approved):

pkg/services/rocketchat/rocketchat_config.go
```diff
@@ -46,7 +47,9 @@ func (config *Config) SetURL(serviceURL *url.URL) error {
 	config.TokenA = path[1]
 	config.TokenB = path[2]
 	if len(path) > 3 {
-		if path[3][0:1] != "@" {
+		if serviceURL.Fragment != "" {
+			config.Channel = "#" + strings.TrimLeft(serviceURL.Fragment, "#")
```
simskij, line 51: https://golang.org/pkg/strings/#TrimLeft

> To remove a prefix, use TrimPrefix instead.

No biggie here as the behavior will be the same, but to keep intentions clear, I think we should switch to `TrimPrefix` anyway.

pkg/services/rocketchat/rocketchat_config.go
```diff
@@ -46,7 +47,9 @@ func (config *Config) SetURL(serviceURL *url.URL) error {
...
+		} else if path[3][0:1] != "@" {
```
simskij, line 52:
```suggestion
		} else if !strings.HasPrefix(path[3], "@") {
```

simskij (changes requested):

simskij:
Thank you for your excellent contribution! Some minor nitpicks above, and then I think we're good to go 👍

pkg/services/rocketchat/rocketchat_config.go
ellisab: Implemented the suggested changes

pkg/services/rocketchat/rocketchat_config.go
ellisab: Implemented the suggested change

simskij:
@all-contributors add @ellisab for code

Issue #76, opened 2020-11-14
Feature request to add custom ports in the rawURL syntax for RocketChat


atighineanu:
By default, send is using the formatted hook, like this "https://%s/hooks/%s/%s", host, tokenA, tokenB. However, in cases when my 443 port is busy with a different service, and I've set the chat service to use port 5000 or 3000, adding the port into the rawURL - it ends up with an error. 

It would be a nice feature to have.

Issue #71, opened 2020-11-05
Rocketchat service segfaults when connection to the server fails


atighineanu:
By default, **send** is using the formatted hook, like this `"https://%s/hooks/%s/%s", host, tokenA, tokenB`. However, in cases when my 443 port is busy with a different service, and I've set the chat service to use port 5000 or 3000, adding the port into the rawURL - it ends up with an error:
```
Imba-MacBookPro:shoutrrr alexeitighineanu$ go run cli/main.go send --url "rocketchat://shoutrrr@<myhost>:3000/<mytokenA>/<mytokenB>/general" --message "heyheyhey"

[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x13869ea]

goroutine 1 [running]:
github.com/containrrr/shoutrrr/pkg/services/rocketchat.(*Service).Send(0xc0000b8980, 0x7ffeefbff721, 0x9, 0xc0000c0088, 0xc0000b8980, 0x0)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/pkg/services/rocketchat/rocketchat.go:41 +0x2ea
github.com/containrrr/shoutrrr.Send(0x7ffeefbff6a4, 0x72, 0x7ffeefbff721, 0x9, 0x9, 0x0)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/shoutrrr.go:24 +0xd7
github.com/containrrr/shoutrrr/cli/cmd/send.Run(0x1913580, 0xc0000dce00, 0x0, 0x4)
        /Users/alexeitighineanu/go/src/github.com/shoutrrr/cli/cmd/send/send.go:49 +0x146
github.com/spf13/cobra.(*Command).execute(0x1913580, 0xc0000dcdc0, 0x4, 0x4, 0x1913580, 0xc0000dcdc0)
        /Users/alexeitighineanu/go/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:842 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x19132e0, 0x0, 0x0, 0x0)
        /Users/alexeitighineanu/go/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:943 +0x336
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/alexeitighineanu/go/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:883
main.main()
[...]

PR #80, opened 2020-11-15
added custom port option for rocketchat

atighineanu:
Closes: https://github.com/containrrr/shoutrrr/issues/76
Closes: https://github.com/containrrr/shoutrrr/issues/71
Added custom port option for --url rocketchat://Host:Port/....

### DIFF
--- a/pkg/services/rocketchat/rocketchat.go
+++ b/pkg/services/rocketchat/rocketchat.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"net/http"
 	"net/url"
-
 	"github.com/containrrr/shoutrrr/pkg/services/standard"
 	"github.com/containrrr/shoutrrr/pkg/types"
 )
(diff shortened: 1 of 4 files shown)


pkg/services/rocketchat/rocketchat.go
```diff
@@ -30,18 +29,28 @@ func (service *Service) Initialize(configURL *url.URL, logger *log.Logger) error
...
+	}
 	if res.StatusCode != http.StatusOK {
 		return fmt.Errorf("failed to send notification to service, response status code %s", res.Status)
 	}
 	return err
 }
 
 func buildURL(config *Config) string {
-	return fmt.Sprintf("https://%s/hooks/%s/%s", config.Host, config.TokenA, config.TokenB)
+	if config.Port != "443" && config.Port != "" {
```
simskij, line 49:
```suggestion
	if config.Port != "" {
```

No need to do an explicit check for 443. It shouldn't matter if it gets appended or not given that `https` is 443 anyway.

simskij (changes requested):
Could you please add some tests for this as well? One where Port is set, and one where it's not.

pkg/services/rocketchat/rocketchat.go
atighineanu: DONE.

simskij (approved):
lgtm. thanks again for contributing!

PR #83, opened 2020-11-15
Added few testcases for rawURL passed as path/#####channel  and path/#channel (used to segfault)

atighineanu:
related to: https://github.com/containrrr/shoutrrr/issues/70
             and: https://github.com/containrrr/shoutrrr/pull/74

### DIFF
--- a/pkg/services/rocketchat/rocketchat_test.go
+++ b/pkg/services/rocketchat/rocketchat_test.go
@@ -172,11 +172,11 @@ var _ = Describe("the rocketchat service", func() {
 			})
 		})
 		When("sending to an URL with badly syntaxed #channel name", func() {
-			rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/###########################testChannel")
-			config := &Config{}
-			config.SetURL(rocketchatURL)
 			It("should properly parse the Channel", func() {
-				Expect(config.Channel).To(ContainSubstring("#testChannel"))
+                rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/###########################testChannel")
+			    config := &Config{}
+  			    config.SetURL(rocketchatURL)
+				Expect(config.Channel).To(ContainSubstring("/###########################testChannel"))
 			})
 			It("should properly parse the Channel", func() {
 			    rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/#testChannel")
--- a/pkg/services/rocketchat/rocketchat_test.go
+++ b/pkg/services/rocketchat/rocketchat_test.go
@@ -178,10 +178,10 @@ var _ = Describe("the rocketchat service", func() {
 			It("should properly parse the Channel", func() {
 				Expect(config.Channel).To(ContainSubstring("#testChannel"))
 			})
-			rocketchatURL, _ = url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/#testChannel")
-			config = &Config{}
-			config.SetURL(rocketchatURL)
 			It("should properly parse the Channel", func() {
+			    rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/#testChannel")
+			    config := &Config{}
+  			    config.SetURL(rocketchatURL)
 				Expect(config.Channel).To(ContainSubstring("#testChannel"))
 			})
 		})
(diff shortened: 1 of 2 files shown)


pkg/services/rocketchat/rocketchat_test.go
```diff
@@ -171,5 +171,19 @@ var _ = Describe("the rocketchat service", func() {
 				Expect(hookURL).To(ContainSubstring("my-domain.com:5055"))
 			})
 		})
+		When("sending to an URL with badly syntaxed #channel name", func() {
+			rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/###########################testChannel")
```
piksel, line 175: I'm actually surprised that it parsed correctly. I wouldn't go out of my way to change the behaviour, but I really don't think that should be a valid URL. It feels like the situation where throwing an error would probably be more helpful instead of just going ahead and assuming that this is what the user intended.

atighineanu: Should it be a fatal or a warning?

simskij: `TrimPrefix` will trim all occurrences of the pattern until it no longer matches.

We don't want to disrupt the user if we can make a reasonable assumption and correct, so a warning is OK in my book.

If the user is indeed trying to do something weird, which I doubt will be the case very often, their send will fail anyway at a later step.

My recommendation would be to just print out that the user needs to look at their config as a warning, strip it, and continue, but @piksel, feel free to decide on this one.

piksel: Yeah, that was basically my takeaway too. If this happens, something weird is going on, but I also think it's kind of unlikely to actually occur, hence why writing code to handle the case might not be worth the effort.
I think my main issue with it was that it was a test explicitly expecting this result (and indeed establishing it as the correct behaviour).
But yeah, if we instead logged a warning, then the test would make much more sense in my opinion. It would then be acknowledged as an incorrect URL, but one we explicitly "correct".
Also, is hashes not allowed in the names of the channels? I remember the IRC channels which actually contained a hash in their name and you would join with "##"-prefix.

ellisab: I checked and hashes are allowed in channel names in rocketchat. I can create a PR and remove strings.TrimPrefix(serviceURL.Fragment, "#") from
[https://github.com/containrrr/shoutrrr/blob/master/pkg/services/rocketchat/rocketchat_config.go#L53](https://github.com/containrrr/shoutrrr/blob/master/pkg/services/rocketchat/rocketchat_config.go#L53)
@piksel @simskij what do you think?

simskij: If there are no constraints regarding hashes in channel names in the rocketchat side of things, I'm all for removing it.

piksel: Agreed

pkg/services/rocketchat/rocketchat_test.go
```diff
@@ -171,5 +171,19 @@ var _ = Describe("the rocketchat service", func() {
...
+			config := &Config{}
+			config.SetURL(rocketchatURL)
+			It("should properly parse the Channel", func() {
+				Expect(config.Channel).To(ContainSubstring("#testChannel"))
+			})
```
simskij, line 180:
```suggestion
			It("should properly parse the Channel", func() {
                rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/###########################testChannel")
			    config := &Config{}
  			    config.SetURL(rocketchatURL)
				Expect(config.Channel).To(ContainSubstring("/###########################testChannel"))
			})
```

pkg/services/rocketchat/rocketchat_test.go
```diff
@@ -171,5 +171,19 @@ var _ = Describe("the rocketchat service", func() {
...
+			rocketchatURL, _ = url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/#testChannel")
+			config = &Config{}
+			config.SetURL(rocketchatURL)
+			It("should properly parse the Channel", func() {
+				Expect(config.Channel).To(ContainSubstring("#testChannel"))
+			})
```
simskij, line 186:
```suggestion
			It("should properly parse the Channel", func() {
			    rocketchatURL, _ := url.Parse("rocketchat://testUserName@rocketchat.my-domain.com:5055/tokenA/tokenB/#testChannel")
			    config := &Config{}
  			    config.SetURL(rocketchatURL)
				Expect(config.Channel).To(ContainSubstring("#testChannel"))
			})
```

simskij (changes requested):
Added some suggestions to reflect what we discussed in the previous review, as well as some partitioning changes to have it match existing test cases and ginkgo conventions. Other than that, I think we're good to go! Thanks again! 🙏

simskij:
👍 🎉

PR #92, opened 2020-12-05
feat(services): add tag-based query/params helper

### LINKED ISSUE #78: Basic generator doesn't work with all current services
I was not able to generate a configuration from cli input.
First, -u or --url cannot be passed as a flag (as described in the docs, see https://github.com/containrrr/shoutrrr/issues/77)
Second, if I use generate -s <service> (even with  -p key=value), it prompts me into something difficult to follow.
```
user@localhost:~/go/src/github.com/atighineanu/shoutrrr> go run cli/main.go generate -s slack
Generating URL for slack using basic generator
Enter the configuration values as prompted

:                                                            <----(*)
: 
BotName: kured-bot
Token: T01EAU1DZ6K                     <-----(**)
panic: reflect: Elem of invalid type slack.Token

goroutine 1 [running]:
reflect.(*rtype).Elem(0x963460, 0x19, 0x199)
        /home/user/go/src/reflect/type.go:918 +0x1a5
github.com/containrrr/shoutrrr/pkg/generators/basic.(*Generator).Generate(0xd43360, 0xa50920, 0xc00000e920, 0xc000095fb0, 0xc00000e8a0, 0x0, 0x2, 0x0, 0x7, 0xc0000975a8, ...)
        /home/user/go/src/github.com/atighineanu/shoutrrr/pkg/generators/basic/basic.go:115 +0xa06
github.com/containrrr/shoutrrr/cli/cmd/generate.Run(0xd08ee0, 0xc00000e8a0, 0x0, 0x2)
        /home/user/go/src/github.com/atighineanu/shoutrrr/cli/cmd/generate/generate.go:106 +0x6d4
github.com/spf13/cobra.(*Command).execute(0xd08ee0, 0xc00000e880, 0x2, 0x2, 0xd08ee0, 0xc00000e880)
        /home/user/go/src/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:842 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xd09180, 0x0, 0x0, 0x0)
        /home/user/go/src/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:943 +0x336
github.com/spf13/cobra.(*Command).Execute(...)
        /home/user/go/src/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:883
main.main()
        /home/user/go/src/github.com/atighineanu/shoutrrr/cli/main.go:9 +0x31
exit status 2
```

* - first two fields of the slice fields from here: https://github.com/containrrr/shoutrrr/blob/ad97e5d949af14a10da9649c1ec03b6dd9b6c628/pkg/generators/basic/basic.go#L30 are completely empty. 
I would assume they are supposed to be the property key:value couple but I am not entirely sure. More description in docs would be welcome. 

** - second, after inputing token from CLI - I always receive the panic error. I've printed the t.Kind() to see what is it in reflect type.go, but it was nil. Tried to put a default and avoid the panic, but ended with (from here https://github.com/containrrr/shoutrrr/blob/ad97e5d949af14a10da9649c1ec03b6dd9b6c628/pkg/generators/basic/basic.go#L126) :
```
Invalid format for field Token: field format is not supported
```
And here I was lost because I have no idea of the package reflect. :)

piksel:
This adds a `format.PropKeyResolver` helper that services can use to automatically implement the query/params serialization based on the config `key` tags.
Example:
```go

import (
        "github.com/containrrr/shoutrrr/pkg/format"
)

type Config struct {
	Title     string   `key:"subject" default:"Shoutrrr Notification"`
}

// Public version of GetURL that creates a new PropKeyResolver when accessed from outside the package
func (config *Config) GetURL() *url.URL {
	resolver := format.NewPropKeyResolver(config)
	return config.getURL(&resolver)
}

// Private version of GetURL that can use an instance of PropKeyResolver instead of rebuilding it's model from reflection
func (config *Config) getURL(resolver types.ConfigQueryResolver) *url.URL {
	return &url.URL{
		Host:       "example",
		Scheme:     Scheme,
                // format.BuildQuery now requires a different interface,
                // which could either be a PropKeyResolver or a Config that implements the old ServiceConfig interface
		RawQuery:   format.BuildQuery(resolver),
	}
}

func (service *Service) Send(message string, params *types.Params) error {
        // Create a copy of the config this send
	config := service.config

        // Mutate the config with overrides from params
	if err := service.propKeyResolver.UpdateConfigFromParams(&config, params); err != nil {
		return err
	} else {
		return service.doSend(client, message, &config)
	}
}
```
Almost all services that uses query fields has been upgraded to use `PropKeyResolver`, and should automatically override the configuration keys when props are passed to the `Send` method.
This also means that services config no longer need to implement the `Get(), Set() and QueryFields` methods, so those along with `standard.QuerylessConfig` has been removed.

Fixes #78


atighineanu:
I've built from this PR. Indeed it doesn't  show ":" preceeded by empty fields, but there is still a crash after token input.

piksel:
@atighineanu I guess it's because of array index out of range?
The token needs to be entered as a comma-separated 3-part array ( `aaaaaaaaaaaaa,bbbbbbbbbbbb,cccccccccccc` ), there is nothing that tells the user that, but now at least it's possible to enter it.
I am looking into either using a custom Generator for Slack or make it possible to specify a minimum length for array fields so that the Basic generator handles them instead of crashing...

simskij (approved):

PR #99, opened 2020-12-20
feat(telegram): add more config options and tests

### LINKED ISSUE #98: Disable Telegram Link Previews
**Is your feature request related to a problem? Please describe.**
Lots of providers are moving to ghcr.io for their images and this results in every watchtower Telegram notification getting what is effectively a Github advert at the bottom as it previews "http://ghcr.io". There doesn't appear to be any way to turn this off on the  Telegram client side.

image

**Describe the solution you'd like**
Telegram provide an option in their API to disable link previews: `disable_web_page_preview` See [https://core.telegram.org/bots/api#sendmessage](https://core.telegram.org/bots/api#sendmessage)

**Additional context**
Don't mind if this is an optional thing but there's unlikely to be much value in providing link previews on the docker notifications.

piksel:
Adds three relevant options for the telegram service to be configured using either query variables or params.

Fixes #98: Add `&preview=no` to the query string to disable previews

### DIFF
--- a/pkg/services/telegram/telegram_test.go
+++ b/pkg/services/telegram/telegram_test.go
@@ -2,6 +2,7 @@ package telegram_test
 
 import (
 	"fmt"
+	"github.com/jarcoal/httpmock"
 	"log"
 	"net/url"
 	"os"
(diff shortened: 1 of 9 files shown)


simskij (approved):

pkg/services/telegram/telegram_test.go
```diff
@@ -120,12 +125,33 @@ var _ = Describe("the telegram plugin", func() {
...
 
+	Describe("sending the payload", func() {
+		var err error
+		BeforeEach(func() {
+			httpmock.Activate()
+		})
+		AfterEach(func() {
+			httpmock.DeactivateAndReset()
+		})
+		It("", func() {
```
piksel, line 136: lol, good description

piksel:
```suggestion
		It("should not report an error if the server accepts the payload", func() {
```

PR #104, opened 2020-12-24
fix smtp config clone toAddresses

sentriz:


piksel:
Thanks!

piksel:
@all-contributors add @sentriz for code

PR #107, opened 2020-12-27
feat: add standard title field

### LINKED ISSUE #30: Standardize "title" field for services
For example, pushover allows you to send a title but it looks like we can only send a message body. Here's the apprise example:

```bash
apprise -t 'my title' -b 'my notification body' \
   'mailto://myemail:mypass@gmail.com' \
   'pbul://o.gn5kj6nfhv736I7jC3cj3QLRiyhgl98b'
```

piksel:
This will add support for multiple key names in the `key` tag (separated by commas), which allows all services to use the `title` key as an alias (or as the primary key).
This also adds said alias to the SMTP service for the `subject` field and a `title` argument to the `send` command.

Fixes #30

### DIFF
--- a/cli/cmd/send/send.go
+++ b/cli/cmd/send/send.go
@@ -5,11 +5,13 @@ import (
 	"log"
 	"os"
 
+	"github.com/spf13/cobra"
+
 	cli "github.com/containrrr/shoutrrr/cli/cmd"
 	u "github.com/containrrr/shoutrrr/internal/util"
+	"github.com/containrrr/shoutrrr/pkg/types"
 	"github.com/containrrr/shoutrrr/pkg/router"
 	"github.com/containrrr/shoutrrr/pkg/util"
-	"github.com/spf13/cobra"
 )
 
 // Cmd sends a notification using a service url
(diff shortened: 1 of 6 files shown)


cli/cmd/send/send.go
```diff
@@ -37,6 +41,7 @@ func Run(cmd *cobra.Command, _ []string) {
 
 	urls, _ := cmd.Flags().GetStringSlice("url")
 	message, _ := cmd.Flags().GetString("message")
+	title, _ := cmd.Flags().GetString("title")
```
simskij, line 44: The cyclomatic complexity of `Run` is >10. I suggest refactoring and extracting from it to make it easier to reason about (and test)

simskij (changes requested):
Minor nitpick, otherwise good to go 👍🏼

cli/cmd/send/send.go
piksel: Yeah, I agree, but skipping it in this branch. The plan is to rewrite this command as a whole pretty soon.

PR #113, opened 2021-01-11
feature: add matrix service

### LINKED ISSUE #112: Feature request: Matrix notifications
It would be nice if a Matrix channel could be used as notification target.

I don't actually know what would be required to achieve this, as I'm neither familiar with Matrix internals nor with Go. But as a user of Watchtower, I'd appreciate support for it 🙂

piksel:
Fixes #112

### DIFF
--- a/docs/services/matrix.md
+++ b/docs/services/matrix.md
@@ -1,0 +1,38 @@
+# Matrix
+
+## URL Format
+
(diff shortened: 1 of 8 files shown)


kubax:
Hi,

is there any ETA on this? Some time ago it seems like watchtower added some check on the url and i can't push notifications to my matrix server because of that (previous it worked fine through a slack compatible webproxy bot, but it is not updates in month, and thus not compatible with the new[?] type of slack notifications)

i'd love to have this feature implemented...

PR #115, opened 2021-01-11
fix(pushover): send all devices in single request

piksel:
,..and don't skip sending if `devices` is not provided


whvanderpost:
Hi,
Today I was trying to get Watchtower to also send Pushover notifications with Shoutrrr and noticed that I have to supply the device parameter otherwise no message is being pushed.
Does the change in this code also end up in the Watchtower project or will Watchtower always require a device parameter?

piksel:
It should be included in the next release of watchtower.

PR #119, opened 2021-01-13
fix(slack): use attachments and add title and color fields

piksel:
- Use legacy `attachments` due to the color support
- Use the normal text field as `title`
- Allow color and title to be set using query and params
- Don't override Bot name by default

### DIFF
--- a/pkg/services/slack/slack.go
+++ b/pkg/services/slack/slack.go
@@ -2,7 +2,6 @@ package slack
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


piksel:
Remove `maxlength` check
https://github.com/containrrr/shoutrrr/blob/38a2e57d321b20434595b07beb5e322793a1434b/pkg/services/slack/slack.go#L40-L42
since it's way lower than the actual slack limits, slack automatically truncates too long messages, and because it doesn't help the user in any way since this is post configuration/initialization.

PR #124, opened 2021-01-21
fix: resolve outstanding golint errors

piksel:
Fixes the various errors that show up when running `golint ./...` locally, effectively hiding any relevant problems in the noise.


pkg/services/smtp/smtp.go
```diff
@@ -73,9 +73,9 @@ func (service *Service) Send(message string, params *types.Params) error {
 	config := service.config.Clone()
 	if err := service.propKeyResolver.UpdateConfigFromParams(&config, params); err != nil {
 		return fail(FailApplySendParams, err)
-	} else {
-		return service.doSend(client, message, &config)
 	}
+
+	return service.doSend(client, message, &config)
```
piksel, line 78: thomas

PR #137, opened 2021-02-07
feat: add support for struct and map fields

piksel:
Adds additional configuration field types for service configs.
Ref #73

### DIFF
--- a/pkg/format/format.go
+++ b/pkg/format/format.go
@@ -5,28 +5,12 @@ import (
 	"strings"
 )
 
-// NotifyFormat describes the format used in the notification body
-type NotifyFormat int
-
-const (
-	// Markdown is the default notification format
-	Markdown NotifyFormat = 0
-)
-
 // ParseBool returns true for "1","true","yes" or false for "0","false","no" or defaultValue for any other value
 func ParseBool(value string, defaultValue bool) (parsedValue bool, ok bool) {
 	switch strings.ToLower(value) {
-	case "true":
-		fallthrough
-	case "1":
-		fallthrough
-	case "yes":
+	case "true", "1", "yes":
 		return true, true
-	case "false":
-		fallthrough
-	case "0":
-		fallthrough
-	case "no":
+	case "false", "0", "no":
 		return false, true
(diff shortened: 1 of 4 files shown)


JonasPf:
I ran into some problems with using `gingko` to run the tests:

```
$ ginkgo -r

Summarizing 6 Failures:

[Fail] the format package SetConfigField when formatting stuct values when setting and formatting [It] should format signed integers identical to input
/Users/jonas/projects/blockdaemon/shoutrrr/pkg/format/format_test.go:236

[Fail] the format package SetConfigField when formatting stuct values when setting and formatting [It] should format unsigned integers identical to input
/Users/jonas/projects/blockdaemon/shoutrrr/pkg/format/format_test.go:236

[Fail] the format package SetConfigField when formatting stuct values when setting and formatting [It] should format enums identical to input
/Users/jonas/projects/blockdaemon/shoutrrr/pkg/format/format_test.go:236

[Fail] the format package SetConfigField when formatting stuct values when setting and formatting [It] should format string slices identical to input
/Users/jonas/projects/blockdaemon/shoutrrr/pkg/format/format_test.go:236

[Fail] the format package SetConfigField when formatting stuct values when setting and formatting [It] should format string arrays identical to input
/Users/jonas/projects/blockdaemon/shoutrrr/pkg/format/format_test.go:236

[Fail] the format package SetConfigField when formatting stuct values when setting and formatting [It] should format prop struct slices identical to input
/Users/jonas/projects/blockdaemon/shoutrrr/pkg/format/format_test.go:236
```

Curiously it works when I run them using `go test ./...` (same as the CI). Anything I'm doing wrong with ginkgo?

piksel:
It's due to the color support. When running with `go test` it detects the output as not being a TTY and strips the color codes from the output, but this does not happen with ginkgo. I usually use `go test` to run, so I hadn't noticed this. I should update the tests to always strip the color information (since that is neither that important, nor what the tests are actually trying to verify here).

```
Expected
      <string>: \x1b[94m-45\x1b[0m
  to equal
      <string>: -45
 ```

Issue #86, opened 2020-11-19
Add support for MQTT (Home Assistant)


simskij:
As requested in https://github.com/containrrr/watchtower/issues/618#issuecomment-703123815, we should consider adding support for sending notifications to home assistant. I'm not sure exactly how, or if, this will work, so some investigation will be needed unless someone has a clear idea.

dougle03:
MQTT messaging perhaps?

rirofal:
As mentioned here...

https://github.com/containrrr/watchtower/discussions/676

simskij:
What would be the use case of this? For instance, which lovelace panel would you use to visualize the messages you receive? As a home assistant user myself, I kind of fail to figure out how this would be used. 😅

Morphy99:
> What would be the use case of this? For instance, which lovelace panel would you use to visualize the messages you receive? As a home assistant user myself, I kind of fail to figure out how this would be used.

I use Home Assistant in a docker container and I use Watchtower to update HA and other containers. It would be nice to get notifications when there is an image update available. I use HA mobile companion for my notifications so it would be good to get it through that.

es-arif:
I'm also in the same boat as @Morphy99.
HA user looking at Watchtower.

_I don't think MQTT is the right choice though._

MQTT is no longer built in to Home Assistant core (as far as I recall) and adding another messaging protocol to this project that supports HTTP doesn't make much sense for maintenance, development, and direction (I will assume).

If you wanted notifications from `shoutrrr` to be detected on a micro-sensor that needs MQTT for the bandwidth/size then sure, but seems like everyone here is echoing that we want `Shoutrrr` to reach Home Assistant core for automation and notifications.

[Creating a webhook is trivial](https://www.home-assistant.io/docs/automation/trigger/#webhook-trigger) and pairing that with a quick template to [forward the webhook as an action to notify](https://www.home-assistant.io/docs/automation/action/) is also very simple and versatile.

Once you get this `Shoutrrr` message through to HA, you can chop and change the action to [whatever notification service you please](https://www.home-assistant.io/integrations/#notifications).

If this is an internal service, then by default no authentication is required on the webhook which further simplifies things.

I'm very fresh but I will be looking at how to pass a generic POST for the moment, but if a PR is required I will come back and give that ago.

fanuch:
I'm also in the same boat as @Morphy99.
HA user looking at Watchtower.

_I don't think MQTT is the right choice though._

[...]

PR #149, opened 2021-03-15
feat(service): add mqtt service

Eugeniosales:
This adds a custom mqtt service as notification options that just publishes the message to a mqtt server such as mosquitto. This feature aims to contribute with the [Issue 86](https://github.com/containrrr/shoutrrr/issues/86) and [Issue 710](https://github.com/containrrr/watchtower/issues/710)

```Go
url := "mqtt://example.com:1083?topic=custom_topic"
err := shoutrrr.Send(url, "Message published through mqtt")
```

The TLS connection option still needs to be added and some tests, but we would like to get a feedback and keep working on it. Thanks!


pkg/services/mqtt/mqtt.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return publishMessageToTopic(message, &config)
+}
+
+// Initialize loads ServiceConfig from configURL and sets logger for this Service
+func (service *Service) Initialize(configURL *url.URL, logger *log.Logger) error {
+	service.Logger.SetLogger(logger)
+	service.config = &Config{
+		DisableTLS:    true,
```
piksel, line 46: Don't disable security features per default. It should need to be explicitly done if the user really needs to do so.

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// getTopic is used to return the topic
+func getTopic(r rune) bool {
+    return r == '='
+}
+
+func (config *Config) setURL(resolver types.ConfigQueryResolver, url *url.URL) error {
+
+	u := strings.FieldsFunc(url.String(), Split)
```
piksel, line 64: Don't manually parse the `URL`, use the built in properties instead.
The `url.Host` contains the "hostname:port" already, no need to split it up and then recombine it.

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,85 @@
...
+	topic := strings.FieldsFunc(url.String(), getTopic)
```
piksel, line 65: Since you tagged the topic field with `key:"topic"`, you can just use:
```go
	for key, vals := range url.Query() {
		if err := resolver.Set(key, vals[0]); err != nil {
			return err
		}
	}
```
to map all query params to their corresponding config fields.

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,85 @@
...
+
+	port, err := strconv.ParseInt(u[2], 10, 64)
```
piksel, line 67: The default value for port is tagged as `1883`, but that's not used if no port is specified.

You could automatically set the config fields to the tagged defaults by using:
```go
	if err := service.pkr.SetDefaultProps(service.config); err != nil {
		return err
	}
```
(in the `service.Initialize()` method)

pkg/services/mqtt/mqtt.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// GetConfig returns the Config for the service
+func (service *Service) GetConfig() *Config {
+	return service.config
+}
+
+// Handle Connection Lost
+var connectLostHandler mqtt.ConnectionLostHandler = func(client mqtt.Client, err error) {
+    fmt.Printf("Connect lost: %v", err)
```
piksel, line 66: Don't directly print to stdout. Use the `service.Logf` method to write output to the configured `log.Logger`

piksel (changes requested):
Yeah, there are some things that needs addressing. I know that this isn't the easiest API to implement, so don't be too disheartened😄 

As a global comment, you probably need to run `go fmt` on the code, since it's not formatted.
Also, you do still have a lot of code copied from the telegram service that probably shouldn't be used (I added a comment about it as well, since you are now sending telegram API messages to MQTT)

pkg/services/mqtt/mqtt_json.go
```diff
@@ -0,0 +1,22 @@
...
+
+// SendMessagePayload sets the JSON to be used as a notification payload for the telegram notification service
+type SendMessagePayload struct {
+	Text                string `json:"text"`
+	Topic               string `json:"topic"`
+	ParseMode           string `json:"parse_mode,omitempty"`
+}
+
+// createSendMessagePayload is used to define the payload
+func createSendMessagePayload(message string, topic string, config *Config) SendMessagePayload {
```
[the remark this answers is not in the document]
Eugeniosales: I used the telegram service as a guideline. I intendend to send a json with the message and topic attributes, but Im going to look depper to see what fits the best.

pkg/services/mqtt/mqtt.go
```diff
@@ -0,0 +1,108 @@
...
+	standard.Standard
+	config *Config
+	pkr    format.PropKeyResolver
+}
+
+// Send notification to mqtt
+func (service *Service) Send(message string, params *types.Params) error {
+	if len(message) > maxlength {
+		return errors.New("message exceeds the max length")
+	}
```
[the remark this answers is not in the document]
Eugeniosales: Ok, I'll fix it. Thanks for the advices!

Eugeniosales:
So, in this update, we solved the problems you pointed out, added the TLS option and some docs.

simskij (changes requested):
Could you please complement your implementation with some tests verifying the intended behavior? Thanks

simskij:
I have nothing more to add. @piksel, if you feel that your feedback has been addressed, feel free to merge. Thank you @Mexazonic and @Eugeniosales for your contributions! 🔥

Eugeniosales:
Is there anything more to be addressed, @piksel? If so, we'd like to get a feedback and keep working on it.

piksel:
Sorry, I have just been busy with other projects. Will get to it soon, I promise!

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,146 @@
...
+// SetURL updates a ServiceConfig from a URL representation of it's field values
+func (config *Config) SetURL(url *url.URL) error {
+	resolver := format.NewPropKeyResolver(config)
+	return config.setURL(&resolver, url)
+}
+
+func (config *Config) getURL(resolver types.ConfigQueryResolver) *url.URL {
+
+	return &url.URL{
+		Host:       fmt.Sprintf("%s:%d", config.Host, config.Port),
```
piksel, line 56: (URL creds)
```suggestion
		User:       util.URLUserPassword(config.Username, config.Password),
		Host:       fmt.Sprintf("%s:%d", config.Host, config.Port),
```

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,146 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/url"
+	"strconv"
+
+	"github.com/containrrr/shoutrrr/pkg/format"
+	"github.com/containrrr/shoutrrr/pkg/types"
```
piksel, line 13: (URL creds)
```suggestion
	"github.com/containrrr/shoutrrr/pkg/types"
	"github.com/containrrr/shoutrrr/pkg/util"
```

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// Config for use within the mqtt
+type Config struct {
+	Host       string `key:"host" default:"" desc:"MQTT broker server hostname or IP address"`
+	Port       uint16 `key:"port" default:"8883" desc:"MQTT server port, common ones are 8883, 1883"`
+	Topic      string `key:"topic" default:"" desc:"Topic where the message is sent"`
+	ClientID   string `key:"clientid" default:"" desc:"client's id from the message is sent"`
+	Username   string `key:"username" default:"" desc:"username for auth"`
+	Password   string `key:"password" default:"" desc:"password for auth"`
```
piksel, line 24: (URL creds)
```suggestion
	Username   string `key:"username" default:"" desc:"username for auth"  url:"User"`
	Password   string `key:"password" default:"" desc:"password for auth"  url:"Pass"`
```

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,146 @@
...
+		opts.SetTLSConfig(tlsConfig)
+	}
+
+	return opts
+}
+
+// GetTLSConfig returns the configuration with the certificates for TLS
+func (config *Config) GetTLSConfig() *tls.Config {
+	certpool := x509.NewCertPool()
+	ca, err := ioutil.ReadFile("ca.crt")
```
piksel, line 121: Requiring the files to have this specific name and be in the working directory of the consuming app makes this really hard to use. Ideally, we should add some kind of generic way to add files/blobs to services (mostly for TLS, so perhaps even a specific TLS-cert interface...)
Right now, this should at least be configurable in the `config` struct.

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,146 @@
...
+
+	clientKeyPair, err := tls.LoadX509KeyPair("client.crt", "client.key")
+	if err != nil {
+		panic(err)
+	}
+	return &tls.Config{
+		RootCAs:            certpool,
+		ClientAuth:         tls.NoClientCert,
+		ClientCAs:          nil,
+		InsecureSkipVerify: true,
```
piksel, line 136: What is the point of loading custom certs when verification is disabled? This should absolutely not be the default.

docs/services/mqtt.md
```diff
@@ -0,0 +1,50 @@
...
+  Default: _empty_
+  Aliases: `clientID`
+
+- **Username** - name of the sender to auth  
+  Default: _empty_
+  Aliases: `clientID`
+
+- **Password** - authentication password or hash  
+  Default: _empty_  
+  Aliases: `password`
```
piksel, line 38: This can auto-generated from the config tags
```suggestion
--8<-- "docs/services/mqtt/config.md"
```

docs/services/mqtt.md
```diff
@@ -0,0 +1,50 @@
+# MQTT
+
+## URL Format
+
+_mqtt://**`host`**:**`port`**?topic=**`topic`**_
+
+## Optional parameters
+
+You can optionally specify the **`disableTLS`**, **`clientID`**, **`username`** and **`password`** parameters in the URL:  
+_mqtt://**`host`**:**`port`**?topic=**`topic`**&disableTLS=true&clientID=**`clientID`**&username=**`username`**&password:**`password`**_
```
piksel, line 10: (URL creds)
```suggestion
_mqtt://**`username`**:**`password`**@**`host`**:**`port`**?topic=**`topic`**&disableTLS=true&clientID=**`clientID`**&username=**`username`**&password:**`password`**_
```

pkg/services/mqtt/mqtt_config.go
```diff
@@ -0,0 +1,146 @@
...
+		Scheme:     Scheme,
+		ForceQuery: true,
+		RawQuery:   format.BuildQuery(resolver),
+	}
+
+}
+
+func (config *Config) setURL(resolver types.ConfigQueryResolver, url *url.URL) error {
+
+	config.Host = url.Hostname()
```
piksel, line 66: (URL creds)
```suggestion
	config.Host = url.Hostname()
	config.Username = url.User.Username()
	password, _ := url.User.Password()
	config.Password = password
```

piksel (changes requested):
Nice work! The username and password ought to be specified using URL `user:pass` fields, both for conistency and some planned logging features (hiding credentials by default in logging). I have marked these suggestions with **(URL creds)** so that they can be batch-merged if you want.
There is also some concern regarding the TLS configuration, but it should be fairly easy to fix!

pkg/services/mqtt/mqtt_config.go
piksel: Implemented in #185

lokanx:
Any progress on this?
Would be an awesome feature to have.

simskij:
This has been untouched for so long that I wouldn't want to merge it. If anyone wants to pick it up again, or provide a completely different implementation, feel free to open a new PR. Closing.

PR #169, opened 2021-05-30
Use absolute path for entrypoint in Dockerfile

arnested:
If the entrypoint uses a relative path we are not able to set a different working directory, i.e. in Dockerfiles extending this.


piksel (approved):
I don't think it was intentional anyway.

PR #170, opened 2021-06-01
Rename Hangouts Chat to Google Chat

arnested:
Google rebranded Hangouts Chat as Google Chat more than a year ago.

Add/keep `hangouts` as an alias scheme for `googlechat` for backward compatibility.


piksel:
Yeah, this was well overdue. The backwards compatibility was handled nicely. If it was any other service, the auto-generated docs would need some verification that they work correctly, but ~~hangouts~~ google chat doesn't use them since there are no config basically.

piksel (approved):

arnested:
I was somewhat surprised that backwards compatibility was so easy. It could probably be done in a ton of other ways, but the ease of this appealed to me.

I don't get the part with the auto-generated docs? Do you have an example of service doing that? And is this something I should add for `googlechat` as well?

piksel:
The most obvious is email/smtp:
https://containrrr.dev/shoutrrr/services/email/
source:
https://github.com/containrrr/shoutrrr/blob/main/docs/services/email.md

Since the URL is exactly like the providers URL, but with a custom scheme, I didn't feel like it added anything for hangouts.

Issue #88, opened 2020-11-24
Almost no services implements params overrides


atighineanu:
So, I tried to test the sender creation feature. 
The code looks like this:
```
url := "slack://A/B/C"
	sender, err := shoutrrr.CreateSender(url)
	if err != nil {
		fmt.Printf("%v", err)
	}
	var hej_da types.Params
	hej_da = make(map[string]string)
	hej_da["channel"] = "random"
	sender.Send("tralala", &hej_da)
```
NOTE: I have created the "random" channel and added the bot to it. Still, when adding the params to the sender, it still didn't send the message to the desired "random" channel. Instead, it sent it to the one corresponding to one of the tokens (I guess it must be C, not sure).

piksel:
Yeah, the `Params` overrides need to implemented in every service, and I think `SMTP` might be the only service that fully implements it.
Slack, like you used in your code above, does not handle them at all:
https://github.com/containrrr/shoutrrr/blob/78573859356243fb2ba448dd9331fa6a06baafda/pkg/services/slack/slack.go#L27

There should be a generic solution for all services to use by just adding a tag to their config.

PR #173, opened 2021-06-09
Slack: add properties for overriding channel and emoji

evenh:
After looking into the new kured 1.7.0 release, I came across a couple of properties that I miss after the transition to shoutrrr. This PR introduces two new properties in the Slack service:

- *channel:* allows for overriding which channel to be notified (default is using the channel specifying in the Slack webhook configuration)
- *emoji:* use a custom icon for the messages posted by shoutrrr

Related to #88, https://github.com/weaveworks/kured/pull/368

### DIFF
--- a/docs/services/slack.md
+++ b/docs/services/slack.md
@@ -19,15 +19,16 @@ Shoutrrr URL:
     
 --8<-- "docs/services/slack/config.md"
 
-!!! info "Color format"
+!!! info "Some properties require URL encoding"
     The format for the `Color` prop follows the [slack docs](https://api.slack.com/reference/messaging/attachments#fields)
     but `#` needs to be escaped as `%23` when passed in a URL.  
     So <span style="background:#ff8000;width:.9em;height:.9em;display:inline-block;vertical-align:middle"></span><code>#ff8000</code> would be `%23ff8000` etc.
-
+    
+    The same applies to the `Channel` property when prefixing with `#` (not required).
 ## Examples
 
 !!! example
     All fields set:
     ```uri
-    slack://ShoutrrrBot@T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX?color=good&title=Great+News
+    slack://ShoutrrrBot@T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX?channel=non-default-chan&color=good&emoji=sunny&title=Great+News
     ```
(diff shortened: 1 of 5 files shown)


piksel:
From the Slack incoming webhook documentation:
> You cannot override the default channel (chosen by the user who installed your app), username, or icon when you're using Incoming Webhooks to post messages. Instead, these values will always inherit from the associated Slack app configuration.

evenh:
That sounds like a retrofitted limitation on Slack's part. The existing implementation already support custom usernames. My patch works in practice and this approach has worked for kured [for two years](https://github.com/weaveworks/kured/blame/807b727ab3a634607ba5d467e46120c7b10584e1/pkg/notifications/slack/slack.go#L18) 🤔

piksel:
It doesn't work with the Shoutrrr test app. My guess is that they stopped allowing it for newly created apps. I disabled it since it was officially not supposed to work and we cannot test it.

evenh:
Huh, that doesn't match my experience (in practice). I have tested against an existing workspace while working on this PR.

For verification, I just created a new Slack workspace (`shoutrrr-june-2021-repro`/`shoutrrrjune2-t6w4466.slack.com`) where I added a new *Incoming WebHooks* integration (marked as legacy). This branch let's me invoke posting to custom channels and using an emoji.

### Default
`./shoutrrr send -v --message=It works --url=slack://xxx/xxx/xxx`
Screenshot 2021-06-10 at 09 57 32

### Overridden emoji and channel name
`./shoutrrr send -v --message=It works --url=slack://xxx/xxx/xxx?emoji=:sunny:&channel=debugging`
Screenshot 2021-06-10 at 09 58 19

piksel:
Ah, the legacy integration. That's probably why. I guess we could add a disclaimer that those options only works with the legacy webhooks.

piksel:
@vsabella Are you referring to apps' workspace bot user oauth tokens as mentioned in the [app authentication basics](https://api.slack.com/authentication/basics#installing)?

That would indeed allow us to use the [`chat.postMessage`](https://api.slack.com/methods/chat.postMessage) endpoint instead which seems to allow all the necessary fields.
It could be reasonably easy to differentiate between the two types of tokens, since the webhook "id" is given in the `Txxxx/Bxxxxx/xxxxx` form and the token is `xoxb-xxxx-xxx-xxxx`.

We could use the API instead of the webhook when given a URL like:
```
slack://bot-user:xoxb-xxxx-xxx-xxxx@api?emoji=sunny&channel=bot-spam
```
My only concern is that it might introduce some confusion for the users ragarding how to set it up.

piksel:
Superseded by #179

PR #182, opened 2021-07-06
feat(slack): add thread/reply support

parrasajad:
- Added `thread_ts` field to Config and JSON 
- If thread  time stamp is supplied it will be sent as reply to message in a thread
- Reference https://api.slack.com/messaging/webhooks#threads


piksel (approved):
LGTM!

I will add this to #179 as well, perhaps also adding logging of the TS of created messages to make it easier to use?

PR #185, opened 2021-07-27
feat: add support for custom root CAs

piksel:
Custom CA support added to the common webclient (`pkg/common/webclient`), which is now used by `slack`, `telegram`, `pushbullet` and `gotify`.

Additionally, the SMTP service got a custom implementation.

Adapting the rest of the services to use `webclient` should be really easy and can be made separately (although releasing with this only supported by *some* services would need a lot of documentation updates).

Needs integration testing (and more unit tests) and a `Router` API.


pkg/common/webclient/client.go
```diff
@@ -0,0 +1,140 @@
...
+func GetJson(url string, response interface{}) error {
+	return DefaultJsonClient.Get(url, response)
+}
+
+// PostJson sends request as JSON and unmarshals the response JSON into the supplied struct using DefaultJsonClient
+func PostJson(url string, request interface{}, response interface{}) error {
+	return DefaultJsonClient.Post(url, request, response)
+}
+
+// WebClient is a JSON wrapper around http.WebClient
```
piksel, line 24: Outdated comment

PR #190, opened 2021-08-10
test(telegram): add generator test

piksel:
adds simple simulation of user interaction with telegram generator as a test


piksel:
This is so stupid. The only added non-test code is the ability to replace the default I/O with testable ones. So the "regular" non-test code is obviously not being hit by tests...

PR #206, opened 2021-10-20
wip: initial tinygo support

piksel:


gedw99:
Hey @piksel 

i guess this PR got abandoned ?

this might sound like a dumb question but what part of the topology does this PR intend to get working in wasm / wasi ?

i have had great success with Wazero. It’s 100% golang with no CGO fun and games. 

 I think it’s an excellent basis for what I presume this PR is looking to do. 
 
I can send example if you want

piksel:
I think that does the opposite of what I wanted to do. The main reason for trying to get it to build with tiny go was to run the URL generation/validation in the browser using WASM. Instead of doing it this way, the current plan is to replace all the reflect code with code generation. That way, we could just as easily generate JavaScript code from the same spec...

PR #215, opened 2021-11-23
feat: send multiple messages when exceeding limits

piksel:
Currently, the discord service sends notifications as one message with several embeds (one for each line). When there are more lines than the number of embeds permitted by the Discord API, they are simply omitted.
This will change that behaviour, so that it will instead send multiple messages when the limit is reached.

Ref https://github.com/containrrr/watchtower/discussions/1126

### DIFF
--- a/pkg/util/partition_message_test.go
+++ b/pkg/util/partition_message_test.go
@@ -1,10 +1,12 @@
 package util
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/containrrr/shoutrrr/pkg/types"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"strings"
 )
 
 var _ = Describe("Partition Message", func() {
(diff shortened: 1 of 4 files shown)


pkg/util/partition_message_test.go
```diff
@@ -99,12 +87,20 @@ func testMessageItemsFromLines(hundreds int, limits types.MessageLimit, repeat i
...
+	// }
 
-	maxChunkSize := Min(limits.ChunkSize, repeat*100)
+	//expectedTotalChunkCount := hundreds / repeat
+	//expectedChunkCountInFirstBatch := Min(limits.TotalChunkSize/maxChunkSize, Min(expectedTotalChunkCount, limits.ChunkCount))
+	// Expect(len(batches[0])).To(Equal(expectedChunkCountInFirstBatch), "Chunk count in first batch")
 
-	expectedChunkCount := Min(limits.TotalChunkSize/maxChunkSize, Min(hundreds/repeat, limits.ChunkCount-1))
-	Expect(len(items)).To(Equal(expectedChunkCount), "Chunk count")
+	// Expect(totalChunks).To(Equal(expectedTotalChunkCount), "Total chunk count")
```
simskij, line 103: Please remove or uncomment before merging.

simskij (changes requested):

pkg/util/partition_message_test.go
piksel: Yep. This was a test that doesn't make any sense anymore, but still needs to be replaced.

Issue #208, opened 2021-10-25
Any way to send pings (http requests) to healthchecks.io


milindpatel63:
I wanna monitor watchtower updates via healthchecks.io
Is there a way to use shoutrrr to send a simple http request when watchtower updates are run.....
i saw shoutrrr has a generic webhook option....can that be used to send a http request....i didn't find any documentation to do so..

simskij:
Hi,

This functionality is not yet released in watchtower, but should be possible to use for what you are describing once it hits `:latest`. The generic service is consumed using the following URI format: `generic+https://`url``

PR #218, opened 2021-12-17
Documenting the Service Schema for Generic

fanuch:
## PR for documenting explicitly the service schema for Generic

Took me about an hour to work out how to use `./shoutrrr send` for a generic request till I read through [the PR for generic](https://github.com/containrrr/shoutrrr/pull/144/files).
Hope this small change can help others.

I wanted to make changes the [generic service document](https://containrrr.dev/shoutrrr/v0.5/services/generic/) but I am not sure of what voodoo is going on as the linked file and directory at `docs/services/generic/config.md` does not exist in the repo!

### DIFF
--- a/docs/services/overview.md
+++ b/docs/services/overview.md
@@ -26,5 +26,5 @@ Click on the service for a more thorough explanation. <!-- @formatter:off -->
 | Service                           | Description                                                                                                                                     |
 | --------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
 | [Logger](./logger.md)             | Writes notification to a configured go `log.Logger`                                                                                             |
-| [Generic Webhook](./generic.md)   | Sends notifications directly to a webhook                                                                                                       |
+| [Generic Webhook](./generic.md)   | *generic+https://`url`* to sends notifications directly to a webhook                                                                                                   |
 
--- a/docs/services/overview.md
+++ b/docs/services/overview.md
@@ -26,5 +26,5 @@ Click on the service for a more thorough explanation. <!-- @formatter:off -->
 | Service                           | Description                                                                                                                                     |
 | --------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
 | [Logger](./logger.md)             | Writes notification to a configured go `log.Logger`                                                                                             |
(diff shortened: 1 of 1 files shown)


piksel:
Hello! The documentation is located in https://github.com/containrrr/shoutrrr/blob/main/docs/services/generic.md. Just put any additional content below the include line. The config.md file is generated from the actual config parser when building the docs (through `generate-service-config-docs.sh`).

fanuch:
> The config.md file is generated from the actual config parser when building the docs

Ah that's brilliant.
I'll add more commits to help the `generic` schema looks closer to the other pages - may as well if I am adding this change in

jbishop129:
Hello, I'm wondering if the `generic` service was documented at some point? it's pretty sparse on the docs section of the website, and nothing of interest in the `generic.md`. I'm asking because I'm struggling to figure out how to get a generic webhook to use `http` instead of always defaulting to `https`. Anway, link to the above documentation file is just showing this:
image

muellermartin:
@jbishop129 I had the exact same issue with `http` instead of `https` URLs causing the error `http: server gave HTTP response to HTTPS client` and via [this comment](https://github.com/containrrr/shoutrrr/issues/208#issuecomment-997075795) in #208 I could correctly guess that the URL must start with `generic+http://` for an HTTP URL.

piksel:
> @jbishop129 I had the exact same issue with `http` instead of `https` URLs causing the error `http: server gave HTTP response to HTTPS client` and via [this comment](https://github.com/containrrr/shoutrrr/issues/208#issuecomment-997075795) in #208 I could correctly guess that the URL must start with `generic+http://` for an HTTP URL.

Yeah, sorry about that. It's `generic+<URL>` where `<URL>` can be either a http:// or https:// URL. All query params will be forwarded, except for the ones listed in the (very limited) [docs](https://containrrr.dev/shoutrrr/v0.5/services/generic/).

If you need to add one of the reserved query params, you can escape it by prefixing it with `__`:

```
generic+https://server/path?__contentType=text/plain
```
That would cause the service to do a POST request, with the `ContentType` header set to `application/json`, to:
```
https://server/path?contentType=text/plain
```

but unescaped

```
generic+https://server/path?contentType=text/plain
```
it would instead send it with the `ContentType` header set to `text/plain`, to:
```
https://server/path
```

PR #220, opened 2021-12-26
Move CLI command to /shoutrrr

arnested:
This will make the CLI command installable by running:

```
go install github.com/containrrr/shoutrrr/shoutrrr@latest
```

When installing using `go install` previously the command got installed
with the binary name `cli`.


simskij (approved):

piksel:
I would probably prefer it to be shoutrrr/shoutrrr instead. I feel like the extra "cmd" doesn't serve any purpose and is less often used as a pattern for CLIs (in my experience).

simskij:
> I would probably prefer it to be shoutrrr/shoutrrr instead. I feel like the extra "cmd" doesn't serve any purpose and is less often used as a pattern for CLIs (in my experience).

It is a common convention for go applications, however.

arnested:
> I would probably prefer it to be shoutrrr/shoutrrr instead. I feel like the extra "cmd" doesn't serve any purpose and is less often used as a pattern for CLIs (in my experience).

I would also say it's quite common. A few examples:

* github.com/go-delve/delve/cmd/dlv
* github.com/fzipp/gocyclo/cmd/gocyclo
* github.com/golangci/golangci-lint/cmd/golangci-lint
* golang.org/x/exp/cmd/gorelease
* golang.org/x/tools/cmd/stringer
* golang.org/x/tools/cmd/goimports

piksel:
Yeah, but those either have multiple binaries (or plan to?), and then it makes sense, but for a single-binary project like this (which instead uses sub-commands), I prefer the style of:
- github.com/onsi/ginkgo/ginkgo
- github.com/spf13/cobra/cobra
- github.com/axw/gocov/gocov
- github.com/davecgh/go-spew/spew

simskij:
You battle it out 😂 I'm down with whichever, although I have a slight preference for the `./cmd` pattern.

arnested:
I'll change it. The most important part for me was just being able to `go install` with the right binary name. I also have preference for `cmd` and think it also communicates the intention clearing when browsing the files and directories. But I can live with both.

arnested:
Change pushed. I'll squash the two commits before this is final and ready to merge.

Putting the command in the folder named `./shoutrrr` required small changes to the build script and Dockerfile because the previous build script produced the cli command binary in the root folder under that name.

piksel:
Haha, sorry for being unreasonable :D I don't care that much either. I'm just selfishly plotting to reduce the huge amount of characters I have to type :D

piksel (approved):

arnested:
I have squashed the commits, and it's ready for merge.

arnested:
@piksel or @simskij, would you merge it?

PR #237, opened 2022-05-17
feat(mattermost): add support for icons

JosephKav:
- What your PR contributes
Icon support on MatterMost:
```bash
./shoutrrr send -u 'mattermost://Argus@mattermost.example.io:443/xxxxxxxxxxx?icon_emoji=https%3A%2F%2Fraw.githubusercontent.com%2Fcontainrrr%2Fshoutrrr%2Fmain%2Fdocs%2Fshoutrrr-logotype.png' -m 'Icon Test'
./shoutrrr send -u 'mattermost://Argus@mattermost.example.io:443/xxxxxxxxxxx?icon_emoji=smiley' -m 'Icon Test'
```
image

- Which issues it solves
Moving my [Argus](https://github.com/release-argus/Argus/issues/33) project to use Shoutrrr. Really want to keep icon support!

- Tests that verify the code your contributing
I tried adding tests (and most of this code is copied from the Slack service), but couldn't get them to pass. Every other service that has query params seems to include the `title` param, but that's no supported on MatterMost, so I'm not sure what to do w.r.t this error. Help fixing this would be appreciated!
```bash
• Failure [0.000 seconds]
services
/path/to/shoutrrr/pkg/services/services_test.go:48
  when passed the a title param
  /path/to/shoutrrr/pkg/services/services_test.go:57
    should not throw an error for mattermost [It]
    /path/to/shoutrrr/pkg/services/services_test.go:71

    Unexpected error:
        <*errors.errorString | 0xc00030e1b0>: {
            s: "title is not a valid config key [icon icon_emoji icon_url]",
        }
        title is not a valid config key [icon icon_emoji icon_url]
    occurred

    /path/to/shoutrrr/pkg/services/services_test.go:99
```

- Updates to the documentation
Done? Let me know if you'd like any more, or changes to them. (Hoping the 'The services does not support any query/param props' would go away with this current commit?)

### DIFF
--- a/pkg/services/mattermost/mattermost_config.go
+++ b/pkg/services/mattermost/mattermost_config.go
@@ -2,15 +2,19 @@ package mattermost
 
 import (
 	"errors"
-	"github.com/containrrr/shoutrrr/pkg/services/standard"
 	"net/url"
 	"strings"
+
+	"github.com/containrrr/shoutrrr/pkg/format"
+	"github.com/containrrr/shoutrrr/pkg/services/standard"
+	"github.com/containrrr/shoutrrr/pkg/types"
 )
 
 //Config object holding all information
 type Config struct {
 	standard.EnumlessConfig
 	UserName string `url:"user" optional:"" desc:"Override webhook user"`
+	Icon     string `key:"icon,icon_emoji,icon_url" default:"" optional:"" desc:"Use emoji or URL as icon (based on presence of http(s):// prefix)"`
 	Channel  string `url:"path2" optional:"" desc:"Override webhook channel"`
 	Host     string `url:"host,port" desc:"Mattermost server host"`
 	Token    string `url:"path1" desc:"Webhook token"`
@@ -36,7 +40,12 @@ func (config *Config) GetURL() *url.URL {
 }
 
 // SetURL updates a ServiceConfig from a URL representation of it's field values
-func (config *Config) SetURL(serviceURL *url.URL) error {
+func (config *Config) SetURL(url *url.URL) error {
+	resolver := format.NewPropKeyResolver(config)
+	return config.setURL(&resolver, url)
+}
+
+func (config *Config) setURL(resolver types.ConfigQueryResolver, serviceURL *url.URL) error {
 
 	config.Host = serviceURL.Hostname()
 	if serviceURL.Path == "" || serviceURL.Path == "/" {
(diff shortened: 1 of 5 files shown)


piksel:
> Every other service that has query params seems to include the title param, but that's no supported on MatterMost.

Just add it anyway. The point is that you should be able to pass `title` to all services without checking for support. If they don't support it, they are free to just ignore it.  
So either put it in the config with `key:"title"`, or manually explicitly ignore it.

pkg/services/mattermost/mattermost_config.go
```diff
@@ -2,15 +2,19 @@ package mattermost
...
+	"github.com/containrrr/shoutrrr/pkg/format"
+	"github.com/containrrr/shoutrrr/pkg/services/standard"
+	"github.com/containrrr/shoutrrr/pkg/types"
 )
 
 //Config object holding all information
 type Config struct {
 	standard.EnumlessConfig
 	UserName string `url:"user" optional:"" desc:"Override webhook user"`
+	Icon     string `key:"icon,icon_emoji,icon_url" default:"" optional:"" desc:"Use emoji or URL as icon (based on presence of http(s):// prefix)"`
```
piksel, line 17:
```suggestion
	Icon     string `key:"icon,icon_emoji,icon_url" default:"" optional:"" desc:"Use emoji or URL as icon (based on presence of http(s):// prefix)"`
	Title    string `key:"title" default:"" desc:"Notification title, optionally set by the sender (not used)"`
```

JosephKav:
Thanks for such a quick response and even giving me the commit suggestion for the fix! 😄

JosephKav:
Any tips to get the codecov higher?
The two that show up on the files page are `return err` being missed after the following
`if err := resolver.Set(key, vals[0]); err != nil {`
or
`if err := service.pkr.UpdateConfigFromParams(config, params); err != nil {`
Any tests that'd get into these?

JosephKav:
Could I get some assistance on increasing the CodeCov please?

piksel:
You can use the generic service compliance tests:
https://github.com/containrrr/shoutrrr/blob/5bca245a82a569f42241b8b000dc7ec220688a8c/pkg/services/bark/bark_test.go#L135-L160

JosephKav:
Hold off on approving this/starting the workflows. Adding more tests and fixing that latest one

pkg/services/mattermost/mattermost_test.go
```diff
@@ -182,6 +264,31 @@ var _ = Describe("the mattermost service", func() {
...
+		})
+		Describe("the service instance", func() {
+			BeforeEach(func() {
+				httpmock.Activate()
+			})
+			AfterEach(func() {
+				httpmock.DeactivateAndReset()
+			})
+			It("should implement basic service API methods correctly", func() {
+				serviceURL := testutils.URLMust("bark://mockhost/mocktoken")
```
piksel, line 288:
```suggestion
				serviceURL := testutils.URLMust("mattermost://mockhost/mocktoken")
```

piksel (approved):
Great work! I changed the scheme in the test URL, it doesn't matter for the test, but it's confusing to have the name of another service in the URL.

PR #241, opened 2022-05-21
fix(slack): limit attachments to 100

### LINKED ISSUE #238: Sending more than ~99 lines to Slack fails with too_many_attachments
# What happens

When sending 99 lines to Slack, shoutrrr splits on newline and successfully sends it as 99 attachments

When sending 100 lines to Slack, shoutrrr splits on newline and attempts to send it as 100 attachments. This fails with the err "failed to send slack notification: webhook response: too_many_attachments"

Note that in some of my other testing, sending 100 lines succeeds but 101 lines fails. I have an off-by-one error somewhere.

# What is expected

Sending any number of lines to Slack, provided it is within the character limit, succeeds.

# Demo

Using the following harness:

```go
package main

import (
    "fmt"
    "github.com/containrrr/shoutrrr"
    "io"
    "os"
)

func main() {
    if len(os.Args) < 2 {
        fmt.Printf("Usage: %s <url>\n", os.Args[0])
        os.Exit(1)
    }

    url := os.Args[1]

    bufB, err := io.ReadAll(os.Stdin)
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }

    buf := string(bufB)

    err = shoutrrr.Send(url, buf)
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }
}
```

Sending 99 lines succeeds:

```plain
% yes | head -n99 | go run shoutrrr_harness.go $URL && echo success
success
```

While sending 100 lines fails:

```plain
% yes | head -n100 | go run shoutrrr_harness.go $URL && echo success
failed to send slack notification: webhook response: too_many_attachments
exit status 1
```

Note that sending 99 relatively long lines still succeeds, indicating this is a problem caused by line count not character count:

```plain
% yes AAAAAAAAAAAAAAAAAAAAAAAAAAA | head -n99 | go run shoutrrr_harness.go $URL && echo success
success
```

piksel:
also removes last attachment if text is empty

fixes #238

### DIFF
--- a/pkg/services/slack/slack_json.go
+++ b/pkg/services/slack/slack_json.go
@@ -73,11 +73,22 @@ type APIResponse struct {
 func CreateJSONPayload(config *Config, message string) interface{} {
 
 	var atts []attachment
-	for _, line := range strings.Split(message, "\n") {
+	for i, line := range strings.Split(message, "\n") {
+		// When 100 attachments have been reached, append the remaining line to the last
+		// attachment to prevent reaching the slack API limit
+		if i >= 100 {
+			atts[len(atts)-1].Text += "\n" + line
+			continue
+		}
 		atts = append(atts, attachment{
 			Text:  line,
 			Color: config.Color,
 		})
+	}
+
+	// Remove last attachment if empty
+	if atts[len(atts)-1].Text == "" {
+		atts = atts[:len(atts)-1]
 	}
 
 	payload := MessagePayload{
--- a/pkg/services/slack/slack_test.go
+++ b/pkg/services/slack/slack_test.go
@@ -2,7 +2,9 @@ package slack_test
 
 import (
 	"errors"
+	"fmt"
 	"log"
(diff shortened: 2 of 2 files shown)


justinsteven:
I'm sorry but I'm not totally familiar with slack's structure and the meaning of attachments. Is it required that attachments be used? I'm guessing it's what's responsible for the pill-shaped graphics beside each line:

image

Perhaps sending more than 100 lines at once isn't a common thing, but having 99 one-line attachments and then a multi-line attachment could look odd.

Basically, do we need to be using attachments, or can we do without them.

piksel:
It's mainly to allow color support, so that you can set the color of the "pill" in front of the text to show errors with red etc.
For some reason, attachments is the only API that still allows that to be specified. We could add another configuration option to send the messages as text instead, but that would instead be limited to the message length limit. Or use the blocks API, which probably have other limits...

Issue #240, opened 2022-05-21
Sending 2000, 4000 or 6000 characters to Discord panics in util.PartitionMessage (index out of range)


justinsteven:
# What happens

Sending exactly 2000, 4000 or 6000 characters to Discord panics

# What should happen

Sending any number of characters to Discord should succeed

# Demo

```go
package main

import (
	"fmt"
	t "github.com/containrrr/shoutrrr/pkg/types"
	"strings"
)

// BEGIN taken from shoutrrr 0.5.3 util and discord packages

var limits = t.MessageLimit{
	ChunkSize:      2000,
	TotalChunkSize: 6000,
	ChunkCount:     10,
}

const maxSearchRunes = 100

// Min returns the smallest of a and b
func Min(a int, b int) int {
	if a < b {
		return a
	}
	return b
}

// PartitionMessage splits a string into chunks that is at most chunkSize runes, it will search the last distance runes
// for a whitespace to make the split appear nicer. It will keep adding chunks until it reaches maxCount chunks, or if
// the total amount of runes in the chunks reach maxTotal.
// The chunks are returned together with the number of omitted runes (that did not fit into the chunks)
func PartitionMessage(input string, limits t.MessageLimit, distance int) (items []t.MessageItem, omitted int) {
	runes := []rune(input)
	chunkOffset := 0
	maxTotal := Min(len(runes), limits.TotalChunkSize)
	maxCount := limits.ChunkCount - 1

	for i := 0; i < maxCount; i++ {
		// If no suitable split point is found, use the chunkSize
		chunkEnd := chunkOffset + limits.ChunkSize
		// ... and start next chunk directly after this one
		nextChunkStart := chunkEnd
		if chunkEnd > maxTotal {
			// The chunk is smaller than the limit, no need to search
			chunkEnd = maxTotal
			nextChunkStart = maxTotal
		} else {
			for r := 0; r < distance; r++ {
				rp := chunkEnd - r
				if runes[rp] == '\n' || runes[rp] == ' ' {
					// Suitable split point found
					chunkEnd = rp
					// Since the split is on a whitespace, skip it in the next chunk
					nextChunkStart = chunkEnd + 1
					break
				}
			}
		}

		items = append(items, t.MessageItem{
			Text: string(runes[chunkOffset:chunkEnd]),
		})

		chunkOffset = nextChunkStart
		if chunkOffset >= maxTotal {
			break
		}
	}

	return items, len(runes) - chunkOffset
}

[...]

PR #245, opened 2022-05-22
fix: text partitioning logic

### LINKED ISSUE #244: Sending 1999, 3999 or 5999 characters to Discord panics in util.PartitionMessage (index out of range)
# What happens

Sending exactly 1999, 3999 or 5999 characters to Discord panics. This is probably due to an incomplete fix for #240 

# What should happen

Sending any number of characters to Discord should succeed

# Demo

```go
package main

import (
        "fmt"
        t "github.com/containrrr/shoutrrr/pkg/types"
        "strings"
)

// BEGIN taken from shoutrrr @bafa90a util and discord packages

var limits = t.MessageLimit{
        ChunkSize:      2000,
        TotalChunkSize: 6000,
        ChunkCount:     10,
}

const maxSearchRunes = 100

// Min returns the smallest of a and b
func Min(a int, b int) int {
        if a < b {
                return a
        }
        return b
}

// PartitionMessage splits a string into chunks that is at most chunkSize runes, it will search the last distance runes
// for a whitespace to make the split appear nicer. It will keep adding chunks until it reaches maxCount chunks, or if
// the total amount of runes in the chunks reach maxTotal.
// The chunks are returned together with the number of omitted runes (that did not fit into the chunks)
func PartitionMessage(input string, limits t.MessageLimit, distance int) (items []t.MessageItem, omitted int) {
        runes := []rune(input)
        chunkOffset := 0
        maxTotal := Min(len(runes), limits.TotalChunkSize)
        maxCount := limits.ChunkCount - 1

        if len(input) == 0 {
                // If the message is empty, return an empty array
                omitted = 0
                return
        }

        for i := 0; i < maxCount; i++ {
                // If no suitable split point is found, start next chunk at chunkSize from chunk start
                nextChunkStart := chunkOffset + limits.ChunkSize
                // ... and set the chunk end to the rune before the next chunk
                chunkEnd := nextChunkStart - 1
                if chunkEnd > maxTotal {
                        // The chunk is smaller than the limit, no need to search
                        chunkEnd = maxTotal
                        nextChunkStart = maxTotal
                } else {
                        for r := 0; r < distance; r++ {
                                rp := chunkEnd - r
                                if runes[rp] == '\n' || runes[rp] == ' ' {
                                        // Suitable split point found
                                        chunkEnd = rp
                                        // Since the split is on a whitespace, skip it in the next chunk
                                        nextChunkStart = chunkEnd + 1
                                        break
                                }
                        }
                }

                items = append(items, t.MessageItem{
                        Text: string(runes[chunkOffset:chunkEnd]),
                })

                chunkOffset = nextChunkStart
                if chunkOffset >= maxTotal {
                        break
                }
        }

        return items, len(runes) - chunkOffset
}

// END taken from shoutrrr @bafa90a

func fuzz(length int) {
        defer func() {
                if r := recover(); r != nil {
                        fmt.Printf("Recovered panic when partitioning message of length %d: %s\n", length, r)
                }
        }()
        PartitionMessage(strings.Repeat("A", length), limits, maxSearchRunes)
}

func main() {
        for i := 0; i <= 20000; i++ {
                fuzz(i)
        }
}
```

```plain
% go run shoutrrr_partitionmessage_fuzz.go
Recovered panic when partitioning message of length 1999: runtime error: index out of range [1999] with length 1999
Recovered panic when partitioning message of length 3999: runtime error: index out of range [3999] with length 3999
Recovered panic when partitioning message of length 5999: runtime error: index out of range [5999] with length 5999
```

piksel:
The fix in #240 only addressed a specific issue, regarding repeated sequences of strings with no whitespace that were exact multiples of the maximum allow chunk size. It just moved the problem to another offset instead.

This does instead fix the real underlying issue and fixing one oddity with previous implementation (why a text string that were the same length as the chunk size would not fit inside one chunk).

Additionally, this includes a test iterating through 8000 text lengths and verifying the contents as well as the chunk sizes.
The general problem with the test is that the logic for calculating the expected sizes works the same way as the implementation. But matched with the unaligned limits and iteration, it should verify the logic sufficiently.

Fixes #244.

### DIFF
--- a/pkg/util/partition_message.go
+++ b/pkg/util/partition_message.go
@@ -28,8 +28,8 @@ func PartitionMessage(input string, limits t.MessageLimit, distance int) (items 
 		// If no suitable split point is found, start next chunk at chunkSize from chunk start
 		nextChunkStart := chunkOffset + limits.ChunkSize
 		// ... and set the chunk end to the rune before the next chunk
-		chunkEnd := nextChunkStart - 1
-		if chunkEnd > maxTotal {
+		chunkEnd := nextChunkStart
+		if chunkEnd >= maxTotal {
 			// The chunk is smaller than the limit, no need to search
 			chunkEnd = maxTotal
 			nextChunkStart = maxTotal
--- a/pkg/util/partition_message.go
+++ b/pkg/util/partition_message.go
@@ -25,10 +25,10 @@ func PartitionMessage(input string, limits t.MessageLimit, distance int) (items 
 	}
 
 	for i := 0; i < maxCount; i++ {
-		// If no suitable split point is found, start next chunk at chunkSize from chunk start
-		nextChunkStart := chunkOffset + limits.ChunkSize
-		// ... and set the chunk end to the rune before the next chunk
-		chunkEnd := nextChunkStart
+		// If no suitable split point is found, use the chunkSize
+		chunkEnd := chunkOffset + limits.ChunkSize
+		// ... and start next chunk directly after this one
+		nextChunkStart := chunkEnd
 		if chunkEnd >= maxTotal {
 			// The chunk is smaller than the limit, no need to search
 			chunkEnd = maxTotal
--- a/pkg/util/partition_message_test.go
+++ b/pkg/util/partition_message_test.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"strconv"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


piksel:
There is something odd going on with the coverage. There is one less covered line, but that line has been removed in this PR, so it shouldn't count.

PR #250, opened 2022-05-25
fix(telegram): update docs and generator for private channels

### LINKED ISSUE #249: Telegram private channels cannot be used with @-prefix syntax
The docs say that 'Props can be either supplied using the params argument, or through the URL', but when chats/channel isn't declared in the url, you get an `error initializing router services: no channels defined in config URL`
https://containrrr.dev/shoutrrr/v0.5/services/telegram/#queryparam_props

I wasn't sure how exactly to go about doing this. I'd be able to give other places where this may happen a go if I encounter it again (and a PR for this is done)

Also, I'd recommend adding [this](https://github.com/containrrr/shoutrrr/issues/223#issuecomment-1088874341) comment to the docs as it makes getting the channel ID a lot easier! That 't.me/' method didn't work for me.

piksel:
The current docs regarding telegram channel _names_ only applies to _public_ channels. Addressing private channels is not possible using the described `@channelname` syntax.
This PR fixes the documentation regarding this, and makes the telegram generator display private channels correctly (instead of just calling them `@`). 

Fixes #249

### DIFF
--- a/docs/services/telegram.md
+++ b/docs/services/telegram.md
@@ -15,15 +15,18 @@ Talk to [the botfather](https://core.telegram.org/bots#6-botfather).
 
 The `chats` param consists of one or more `Chat ID`s or `channel name`s. 
 
-### Channels
-The channel names can be retrieved in the telegram client in the `Channel info` section. 
+### Public Channels
+The channel names can be retrieved in the telegram client in the `Channel info` section for public channels. 
 Replace the `t.me/` prefix from the link with a `@`.
 
 !!! note
     Channels names need to be prefixed by `@` to identify them as such.
 
+!!! note
+    If your channel only has an invite link (starting with `t.me/+`), you have to use it's Chat ID (see below)
+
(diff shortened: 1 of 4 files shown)


JosephKav:
I'd still recommend the forward message to 'JSON Dump Bot' method to get the chat ID. Doesn't require a PC/Docker compared to `docker run --rm -it containrrr/shoutrrr generate telegram`
Up to you though of course

piksel:
@JosephKav I added another section, but replace `@JsonDumpBot` with `@UserInfoBot`, which is made by the same author. It's slightly easier to use, but provides the same information.

piksel:
Since neither `@JsonDumpBot`, nor `@UserInfoBot` could help getting the Chat ID for group chats, I wrote a bot specifically for this purpose. It can be reached at [@shoutrrrbot](https://t.me/shoutrrrbot]. Updated the docs to point to it instead.

PR #252, opened 2022-05-26
fix(smtp): use `usestarttls` (match the docs)

JosephKav:
The [smtp docs](https://containrrr.dev/shoutrrr/v0.5/services/smtp/config/#queryparam_props) mention `UseStartTLS`, but it doesn't accept `usestarttls`, it wants `starttls`. As `UseStartTLS` is the Go var, it was probably designed to be `usestarttls` as the key, so here's a PR for that.

Thanks to @samcro1967 for spotting this in the issue for my PR that added Shoutrrr
https://github.com/release-argus/Argus/issues/33#issuecomment-1139113316

Wasn't sure about how/whether you'd want to keep support for `starttls` for a while as it would show in the docs if I put it under keys, right?

### DIFF
--- a/pkg/services/smtp/smtp_config.go
+++ b/pkg/services/smtp/smtp_config.go
@@ -3,9 +3,10 @@ package smtp
 import (
 	"errors"
 	"fmt"
-	"github.com/containrrr/shoutrrr/pkg/util"
 	"net/url"
 	"strconv"
+
+	"github.com/containrrr/shoutrrr/pkg/util"
 
 	"github.com/containrrr/shoutrrr/pkg/format"
 	"github.com/containrrr/shoutrrr/pkg/types"
@@ -23,7 +24,7 @@ type Config struct {
 	Subject     string    `desc:"The subject of the sent mail" key:"subject,title" default:"Shoutrrr Notification"`
 	Auth        authType  `desc:"SMTP authentication method" key:"auth" default:"Unknown"`
 	Encryption  encMethod `desc:"Encryption method" default:"Auto" key:"encryption"`
-	UseStartTLS bool      `desc:"Whether to use StartTLS encryption" default:"Yes" key:"starttls"`
+	UseStartTLS bool      `desc:"Whether to use StartTLS encryption" default:"Yes" key:"usestarttls"`
 	UseHTML     bool      `desc:"Whether the message being sent is in HTML" default:"No" key:"usehtml"`
 }
 
--- a/pkg/services/smtp/smtp_config.go
+++ b/pkg/services/smtp/smtp_config.go
@@ -6,10 +6,9 @@ import (
 	"net/url"
 	"strconv"
 
-	"github.com/containrrr/shoutrrr/pkg/util"
-
 	"github.com/containrrr/shoutrrr/pkg/format"
 	"github.com/containrrr/shoutrrr/pkg/types"
+	"github.com/containrrr/shoutrrr/pkg/util"
 )
 
 // Config is the configuration needed to send e-mail notifications over SMTP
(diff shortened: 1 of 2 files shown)


pkg/services/smtp/smtp_test.go
```diff
@@ -42,9 +43,10 @@ var _ = Describe("the SMTP service", func() {
 	})
 	When("parsing the configuration URL", func() {
 		It("should be identical after de-/serialization", func() {
-			testURL := "smtp://user:password@example.com:2225/?auth=None&encryption=ExplicitTLS&fromaddress=sender%40example.com&fromname=Sender&starttls=No&subject=Subject&toaddresses=rec1%40example.com%2Crec2%40example.com&usehtml=Yes"
+			testURL := "smtp://user:password@example.com:2225/?auth=None&encryption=ExplicitTLS&fromaddress=sender%40example.com&fromname=Sender&subject=Subject&toaddresses=rec1%40example.com%2Crec2%40example.com&usehtml=Yes&usestarttls=No"
```
JosephKav, line 46: Just noticed why I had to move this to the end. It alphabetises the params right?

pkg/services/smtp/smtp_config.go
```diff
@@ -23,7 +24,7 @@ type Config struct {
 	Subject     string    `desc:"The subject of the sent mail" key:"subject,title" default:"Shoutrrr Notification"`
 	Auth        authType  `desc:"SMTP authentication method" key:"auth" default:"Unknown"`
 	Encryption  encMethod `desc:"Encryption method" default:"Auto" key:"encryption"`
-	UseStartTLS bool      `desc:"Whether to use StartTLS encryption" default:"Yes" key:"starttls"`
+	UseStartTLS bool      `desc:"Whether to use StartTLS encryption" default:"Yes" key:"usestarttls"`
```
piksel, line 27: We don't want to make a breaking change, but demoting `starttls` to an alias is fine.
```suggestion
	UseStartTLS bool      `desc:"Whether to use StartTLS encryption" default:"Yes" key:"usestarttls,starttls"`
```

pkg/services/smtp/smtp_config.go
```diff
@@ -3,10 +3,11 @@ package smtp
 import (
 	"errors"
 	"fmt"
-	"github.com/containrrr/shoutrrr/pkg/util"
 	"net/url"
 	"strconv"
 
+	"github.com/containrrr/shoutrrr/pkg/util"
+
```
piksel, line 10: (**very** nit)
```suggestion
```

piksel (changes requested):
Thanks for finding this!

It's an oversight in the docs generation. When the key doesn't match the field name it should make a note of that, instead of just assuming they will be same.

In any case, this looks good, apart from the breaking change, but that's easily fixed.

pkg/services/smtp/smtp_test.go
JosephKav: ah, must be the yaml encoder that alphabetises my maps then. Unless I have been extremely 'lucky' to have it alphabetised correctly every time I checked after a save

pkg/services/smtp/smtp_config.go
JosephKav: Alright, wasn't sure whether you had a way of making it an alias that wouldn't show in the docs for if you ever wanted to remove this alias

pkg/services/smtp/smtp_config.go
JosephKav: np! did it as a separate commit though as `gofmt` alphabetised those impots. Not sure why it didn't merge them originally

pkg/services/smtp/smtp_test.go
piksel: Indeed it does, it has it's own map impl: https://pkg.go.dev/gopkg.in/yaml.v2#MapSlice

piksel (approved):

pkg/services/smtp/smtp_test.go
JosephKav: I wish v3 has a MapSlice so that I could choose the order when I save it. Had to make it save the file, then go through and check whether the blocks have been written in the correct order (they won't have been...), then bubble sort those blocks back to how it should be and save again! I did spend a while on ordering the maps, but then found out they were alphabetised and am calling that good enough

pkg/services/smtp/smtp_config.go
piksel: We might make some breaking changes in a `1.0` release, but until then I see no reason to remove an alias. Users generally just configure once and then expect it to continue working. So that seems like the highest priority.

piksel:
@all-contributors add @JosephKav for bug

PR #274, opened 2022-10-17
Remove duplicate URLs so we don't send to the same service twice

arnested:
It doesn't make much sense to send a message to the same service twice.

It's probably not likely that a user will make the mistake by adding the same service twice by adding it twice on the command
line (`shoutrrr send --url=my://service --url=my://service`), but in a script or Docker setup it could happen accidentally using an environment variable like this:

```bash
export SHOUTRRR_URL=my://service

shoutrrr --url=${SHOUTRRR_URL}
```

This has happened to me in a Docker setup. And I'm not too proud to admit that it took me a long time before I figured out why each message was duplicated...

### DIFF
--- a/shoutrrr/cmd/send/send.go
+++ b/shoutrrr/cmd/send/send.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/containrrr/shoutrrr/internal/dedupe"
 	intutil "github.com/containrrr/shoutrrr/internal/util"
 	"github.com/containrrr/shoutrrr/pkg/router"
 	"github.com/containrrr/shoutrrr/pkg/types"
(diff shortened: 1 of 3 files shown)


shoutrrr/cmd/send/send.go
```diff
@@ -46,6 +47,10 @@ func run(cmd *cobra.Command) error {
 	verbose, _ := flags.GetBool("verbose")
 
 	urls, _ := flags.GetStringArray("url")
+	// Remove duplicate service URLs so we don't send to the same
+	// service twice
+	_ = dedupe.Do(&urls)
```
piksel, line 52:
```suggestion
	urls = utils.RemoveDuplicates(&urls)
```

I think this should not need an external dependency, especially not one that uses reflection (I am trying to get rid of reflection altogether in `next`).
Since we only need to do this for a string slice, it could be implemented as:

```go
func RemoveDuplicates(src []string) []string {
	unique := make([]string, 0, len(urls))

	for _, s := range src {
		found := false
		for _, us := range unique {
			if us == s {
				found = true
				break
			}
		}
		if !found {
			unique = append(unique, u)
		}
	}

	return unique
}
```

arnested: You're right. I was probably too focused on making a minimal diff.

I pushed your implementation together with a test for it.

piksel (approved):
Looks good!

PR #291, opened 2022-11-03
chore(deps): bulk update dependencies

piksel:
Bulk update of dependencies:
- #286 
- ~~#287~~
- #288 
- ~~#289~~
- #290


piksel:
Updating ginkgo to v2.4.0 requires go v1.17, so that update has been reverted (hence why the CI failed on v1.16).

Opened a PR regarding the requirement change in https://github.com/onsi/ginkgo/issues/1073.

PR #304, opened 2022-12-10
Make the cli available as a snap

simskij:
Given how good and easy the CLI has become to use, it only makes sense for it to also be available as a snap on the snapstore. I've gone ahead and registered the snap name already and published a revision on the stable track.

This PR moves the ownership of the snap packing spec to the containrrr team.

[Snapstore Entry](https://snapcraft.io/shoutrrr)


simskij:
In the scope of this PR, Im taking zero responsibility for the codacy check failing. 😅

Issue #293, opened 2022-11-23
Latest version is not latest on the website


AlekSi:
https://containrrr.dev/shoutrrr/v0.6/ gives me:

> You're not viewing the latest version. Click here to go to latest.

Where latest leads to 0.5 docs.

<img width="528" alt="CleanShot 2022-11-23 at 18 56 06@2x" src="https://user-images.githubusercontent.com/11512/203577907-80bd41ae-79c3-4f50-abec-bbc50082d0a4.png">

piksel:
Yes, something is wrong with the docs generation. I will look into it. Thanks for reporting!

andig:
This is still an issue. Selecting 0.7 will warn "not viewing latest release".

piksel:
Yeah, I noticed. I will make a manual fix as soon as I get back.

piksel:
Found the missing piece. This needs to be run whenever the minor changes:
```
mike deploy --push --update-aliases 0.7 latest
```

I will keep this open until it has been added to CI, but at least it's fixed in the current docs.

PR #347, opened 2023-04-29
Update docs.yml

simskij:
Updates CI to update aliases automatically.

resolves https://github.com/containrrr/shoutrrr/issues/293

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -63,4 +63,5 @@ jobs:
 
 
       - name: Deploy docs
-        run: mike deploy --push $DOCS_VERSION
+        run: mike deploy --push --update-aliases $DOCS_VERSION latest
+


.github/workflows/docs.yml
```diff
@@ -63,4 +63,5 @@ jobs:
 
 
       - name: Deploy docs
-        run: mike deploy --push $DOCS_VERSION
+        run: mike deploy --push --update-aliases $DOCS_VERSION latest
```
piksel, line 66: Hmm, I'm not so sure this is correct. The value of `$DOCS_VERSION` is `latest` (when running on that branch). Wouldn't that just alias it to itself?

piksel:
Yeah, I finally found the missing piece of the puzzle. The workflow is different in `latest` since that one is pushing to different versioned tags: https://github.com/containrrr/shoutrrr/blob/latest/.github/workflows/docs.yml#L70

So, the change in this PR, but applied to that file instead should work. Obviously, we should use the same CI workflow for both branches and have if-statements that selects the correct tag instead.

PR #354, opened 2023-05-24
Update matrix docs to mention `title` caveat

### LINKED ISSUE #335: Matrix: Title is not used
I am currently trying to configure watchtower to send messages to matrix via shoutrrr.
I use `&title=xyz` as part of the url as stated in the documentation.
The [documentation for the matrix service](https://containrrr.dev/shoutrrr/v0.6/services/matrix/) states that title is a supported property.

Though I can not find it used at all in the source code for the service.
It seems like it is not implemented as running the service CLI doesn't produce a title either?

```
$ docker run --rm -v $PWD/discord-data:/out -it containrrr/shoutrrr:latest send "matrix://watchtower:<pass>@<server>/?title=123" WTF
Notification sent
```
image

simskij:
Updates the docs to include a note on the caveat with `title` not being used for Matrix messages.

resolves #335

### DIFF
--- a/docs/services/matrix.md
+++ b/docs/services/matrix.md
@@ -1,5 +1,9 @@
 # Matrix
 
+> ### ⚠️ Usage of the `title` parameter
+> Do note that Matrix will discard any information put in the `title` parameter as the service has no analogue to a
+> a title. Instead, use a custom message format to supply your wanted title as part of the message.
+
 ## URL Format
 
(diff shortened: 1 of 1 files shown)


piksel (commented):
The title field is such a wildcard. This note applies to some other services as well. But yeah, thanks for making a PR for the docs.

PR #362, opened 2023-06-11
feat(telegram): add support for message_thread_id

### LINKED ISSUE #361: Support for Telegram thread ID
The `sendMessage` method (https://core.telegram.org/bots/api#sendmessage) now supports an optional `message_thread_id` to send the message to a super group topic. It would be great if this can be supported by shoutrrr.

My preferred way to support it is to specify the chat id of the super group in the format of:

```
telegram://$token@telegram?chats=$chat_id:$message_thread_id,$chat_id2,$chat_id3:$message_thread_id
```

yegle:
Reference: https://core.telegram.org/bots/api#sendmessage

closes #361

### DIFF
--- a/docs/services/telegram.md
+++ b/docs/services/telegram.md
@@ -26,7 +26,7 @@ Replace the `t.me/` prefix from the link with a `@`.
     If your channel only has an invite link (starting with `t.me/+`), you have to use it's Chat ID (see below)
 
 !!! note
-    A `message_thread_id` param ([reference](https://core.telegram.org/bots/api#sendmessage)) can be added, with the format of `$chat_id:$message_thread_id`.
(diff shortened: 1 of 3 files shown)


piksel:
Looks good, but there are no tests. Also, how do you retrieve the thread ID?

yegle:
Test added, info on how to obtain the message_thread_id added.

piksel (approved):
Great work! I'm not sure how I feel about linking a SO question in the docs, but perhaps it's better than nothing. We could always update the docs at a later time.

yegle:
Is it okay to keep the `$chat_id:$message_thread_id` format? I only proposed it because it's easy to implement, but we can change to something more formal.

piksel:
Yeah, I think it's a good way to do it. Normally I would prefer to add another parameter, but since this is directly relating to a target it makes sense to use this. It also doesn't need to be escaped which is a plus!

instantdreams:
Documentation needs to be updated to reflect use and examples of Topic ID

Issue #231, opened 2022-04-02
Add support for Apprise


PiDroid-B:
Hello,

I use Watchtower and i like this amazing project !

To improve the management of migrations, i use Apprise as a "notification broker" then i can change all the end points at the same place (i.e. : migration of mattermost, add discord channel etc... without change it on each services).

I search about integration of Apprise in Watchtower and/or Shoutrrr and found this : https://github.com/containrrr/shoutrrr/issues/59

i like a lot if i can use Apprise as like as i can use Discord or Mattermost (or everything else in this page : https://containrrr.dev/shoutrrr/v0.5/services/overview/ )

There is a futur integration of Apprise with Shoutrrr ?

Thank's a lot !

piksel:
Yeah, we should probably add an `apprise` provider that puts the content in the `body` field. You can already do it by using the `generic` service, but you would have to make a custom template payload, which makes it a lot harder to use.

RichyHBM:
Has there been any more thought in this? Or any guides on how to achieve it using the generic method?

piksel:
@RichyHBM No, not really. I have not had the time to look into it, my statement above is just from a technical stand, it *should* be possible to do using a custom template, but it would require some trial and error. You would need to create a custom template that matches the apprise JSON body, something like:
```json
{
   "urls": {"pbul://apikey, mailto://user:pass@gmail.com"},
   "title": {"{{ .Title }}"},
   "body": {"{{range .Entries -}}{{.Message}}{{\"\n\"}}{{- end -}}"},
   "type": {"info"},
}
```
(this is assuming you are using watchtower)

simskij:
> Has there been any more thought in this? Or any guides on how to achieve it using the generic method?

Feel free to have a stab at it if you want! 👍🏼

TheFiZi:
> Has there been any more thought in this? Or any guides on how to achieve it using the generic method?

A workaround for now based on this PR: https://github.com/containrrr/shoutrrr/pull/364

Set `WATCHTOWER_NOTIFICATION_URL` to: `generic://[APPRISE FQDN]/notify/[CONFIG ID]?template=json&messagekey=body&title=title&tags=[YOUR TAG]`

Example: `generic://apprise.mydomain.com/notify/my_apprise_config_id?template=json&messagekey=body&title=title&tags=discord_notifications`

PR #364, opened 2023-06-13
Add example calling Apprise

igrybkov:
Adding documentation example on how to integrate Shoutrrr with the [Apprise API](https://github.com/caronc/apprise-api).

It helps with #231, although it's not a complete replacement since this example doesn't support tags and custom URLs and only allows integration with an existing Apprise configuration/ endpoint.


bwirt:
+1 for this, was tearing my hair out trying to figure this out.  Would be helpful to have this merged into main docs.  I wanted to use Apprise which is configured to use Outlook SMTP, which shoutrrr doesn't support (would also suggest mentioning this somewhere as I wasted some time troubleshooting it).

igrybkov:
lol I already forgot about this PR. It's been over a year now!

Hello @piksel, can we get this merged to share the knowledge in the docs? Or let me know if you want me to improve this in some way before merging it.

PR #373, opened 2023-07-06
fix(ifttt): treat all 2xx http responses as success

eoleedi:
Originally, shoutrrr sees only http status code 204 as a successful response from IFTTT webhook service.
However, http status code 200 should be also seen has a successful response.

This prevents an incorrect error log from generating.

### DIFF
--- a/pkg/services/ifttt/ifttt.go
+++ b/pkg/services/ifttt/ifttt.go
@@ -72,7 +72,7 @@ func doSend(payload []byte, postURL string) error {
 	if err != nil {
 		return err
 	}
-	if res.StatusCode != http.StatusNoContent {
+	if res.StatusCode != http.StatusNoContent && res.StatusCode != http.StatusOK {
 		return fmt.Errorf("got response status code %s", res.Status)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


piksel:
I tried looking at the documentation to find what status codes are expected, but I couldn't find any explicit documentation. It seems like the service is using this:
https://ifttt.com/maker_webhooks/triggers/event

There is some documentation at https://maker.ifttt.com/use/your-key but it doesn't contain any response statuses. While testing, it does indeed now return `200` with the text body of `Congratulations! You've fired the <EVENT_NAME> event`. Perhaps they changed this behaviour?

eoleedi:
I also didn't find any explicit documentation.
Since it is a service provided by IFTTT, maybe we can assume that the API follows the General API requirements of IFTTT.
https://ifttt.com/docs/api_reference

Dig deeper into the OpenAPI of IFTTT
There is a similar but not the exact same API endpoint, where `200` is used.
https://github.com/IFTTT/service-api/blob/99218c72a9564b842f30e367de0f2fe0f450f3b7/service-api.yaml#L331C1-L365C60

If that's the case, should we keep `204` as successful response just in case?

piksel:
Yeah, any 2xx status denotes a success, so we should never treat those as errors anyway.

PR #405, opened 2023-11-02
feat(#404): implement basic webex functionality to shoutrrr

### LINKED ISSUE #404: Webex support
Support for Cisco Webex would be great. It does have a chat-bot functionality, basic webhooks are possible but there is an issue when you start adding more and more webhooks. You end up with a new version of the person who created the webhook, with a random ID at the end, so a bot token is just better to use and recommend.

(I plan on looking at adding this, some point this weekend hopefully)

---

**Notes:**

Implement basic webex notification using cards: https://developer.webex.com/buttons-and-cards-designer

```
webex://bot-token@room-id
```

Likely copy/paste of discord, where custom JSON payload _can_ be used if wanted. Allowing for more complex card structures if you want, or a basic markdown message.

surdaft:
Implements a basic version of webex, future plans to add cards as well, but for now this satisfies webex being available.

image

closes #404  (heh, 404)

### DIFF
--- a/docs/services/webex.md
+++ b/docs/services/webex.md
@@ -1,0 +1,13 @@
+# Webex
+
+Adds Webex service functionality
+
+!!! info "Your bot must be in the room"
+    Your bot must be invited to the room before a message can be sent by it.
+
+## Examples
+
+!!! example "General"
+    ```uri
+    webex://your-bot-token@your-room-id
+    ```
--- a/docs/services/webex.md
+++ b/docs/services/webex.md
@@ -5,9 +5,13 @@ Adds Webex service functionality
 !!! info "Your bot must be in the room"
     Your bot must be invited to the room before a message can be sent by it.
 
-## Examples
+## URL Format
 
-!!! example "General"
+The URL format requires the `@webex?` to be present. This is to handle the URL
+being parsed properly and is ignored.
+
+!!! info ""
     ```uri
-    webex://your-bot-token@your-room-id
-    ```
+   webex://__`token`__@webex?rooms=__`room-1`__[,__`room-2`__,...]
+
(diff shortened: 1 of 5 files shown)


piksel (commented):
This looks really well done. Good job!

My suggestion would be to pass the roomId as a query parameter instead to allow for sending to multiple rooms. It would require the hostname to be a static string (which is generally just ignored by other services that have this problem). 

Also, even though it's very simple, add the macro for including auto-generated docs so that the docs align with the other services. It's the scissor-looking thing at the bottom of the other services markdown files.

surdaft:
Thanks very much! Query param for roomID, fair enough I will take a poke. Do you have an example other service?

Ooh I see, I wondered what that was for. Thanks for the clarification, I will add the scissors when I have a moment. :)

piksel:
Sure, the telegram service is similar and loops over its targets (`chats`) when sending:
https://github.com/containrrr/shoutrrr/blob/52149dcefc8eef26627cc5f48a0c3997280bc879/pkg/services/telegram/telegram.go#L25

It's configured using the `key` tag (the second arg `channels` is an alias for backwards compatibility):
https://github.com/containrrr/shoutrrr/blob/52149dcefc8eef26627cc5f48a0c3997280bc879/pkg/services/telegram/telegram_config.go#L19

Even if you only support a single room for now, using a query variable allows for adding support for multiple ones in the future without breaking compatibility. It also aligns it better with the other services. Thanks for taking the time!

surdaft:
- I have updated to provide a `rooms` query parameter, which is comma separated like telegram.
- I also tweaked the docs page to better reference how the discord and telegram docs show the URL structure, it was slightly different. I don't remember where I copied it from. 😅
- Snippers added

PR #408, opened 2023-11-05
fix: fix an incorrect printf and add a missing error check

favonia:
This PR fixes a few minor issues I noticed when checking the code.


piksel (approved):
Good catches!

PR #409, opened 2023-11-05
feat: add unique service instance identifiers to router

piksel:

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -3,7 +3,6 @@ package router
 import (
 	"fmt"
 	"net/url"
-	"reflect"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


favonia:
Thank you! In case it helps, it's okay for my application to show duplicate service names. However, I can also see why you want to make them distinct! 😃

favonia:
Alternative approach: maybe in 1.0 API, there could be an API returning the list of services directly, and each service must implement a function showing a name. The ServiceRouter is then closer to a "composite" service that can queue messages and do AsyncSend.

PS: I want to emphasize that I can use the current library as it is just fine, in case you think more discussions could be helpful. :grin:

PR #423, opened 2023-12-31
Mention Green Orb

atgreen:
As long as you are referencing users of shoutrrr, here's a mention of Green Orb.
Thanks for shoutrrr!


piksel (approved):
Looks really useful! Happy to add it.

PR #430, opened 2024-01-19
fix(smtp): use param props from params arg

JosephKav:
was ignoring encryption, subject, fromname and fromaddress

Noticed from https://github.com/release-argus/Argus/issues/352 - Setting fromname in param props that are given with `sender.Send(toSend, params)` was ignored.

### DIFF
--- a/pkg/services/smtp/smtp.go
+++ b/pkg/services/smtp/smtp.go
@@ -71,7 +71,7 @@ func (service *Service) Send(message string, params *types.Params) error {
 		return fail(FailApplySendParams, err)
 	}
 
-	client, err := getClientConnection(service.config)
+	client, err := getClientConnection(&config)
 	if err != nil {
 		return fail(FailGetSMTPClient, err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/services/smtp/smtp.go
```diff
@@ -71,7 +71,7 @@ func (service *Service) Send(message string, params *types.Params) error {
 		return fail(FailApplySendParams, err)
 	}
 
-	client, err := getClientConnection(service.config)
+	client, err := getClientConnection(&config)
```
JosephKav, line 74:
```go
	if useImplicitTLS(config.Encryption, config.Port) {
		conn, err = tls.Dial("tcp", addr, &tls.Config{
			ServerName: config.Host,
		})
	} else {
		conn, err = net.Dial("tcp", addr)
	}
```

This was ignoring the value in the params var

pkg/services/smtp/smtp.go
```diff
@@ -239,7 +237,7 @@ func (service *Service) getHeaders(toAddress string, subject string) map[string]
 	}
 
 	return map[string]string{
-		"Subject":      subject,
+		"Subject":      conf.Subject,
```
JosephKav, line 240: `conf.Subject` may contain changes from the var

PR #449, opened 2024-07-26
Removes leading '$' from bash commands in README.md so they work when copy pasted

popaurel:


arnested (approved):

simskij:
Thank you for your contribution! While your interest in contributing to shoutrrr is much appreciated, the dollar signs were added deliberately.

It’s a bad practice to blindly copy/pasting snippets from the internet in your terminal - having to remove the dollar sign requires the user to at least wake up for long enough to reflect on what they are doing. Closing without action.

Issue #396, opened 2023-10-08
Discord notification not working right without  - no splitlines 


masterwishx:
Related to : 
https://github.com/containrrr/watchtower/discussions/1787#discussioncomment-7217122

`?splitlines=no` at the end of discord notify fix the issue

PR #498, opened 2026-05-18
Discord: handle splitLines when set as message param

danielskowronski:
Solves #396

While `splitLines` was allowed as both message and service parameter, `Send` was calling `CreateItemsFromPlain` based on `service.config.SplitLines`. This meant that per-message setting was impossible and forced users to set it in URI (which requires proper URI parsing).

Introduced change to check if `splitLines` is set as message param and if so, parse as boolean overriding `service.config.SplitLines` for this message only.

CleanShot 2026-05-18 at 20 27 12

### DIFF
--- a/pkg/services/discord/discord.go
+++ b/pkg/services/discord/discord.go
@@ -36,11 +36,17 @@ const (
 func (service *Service) Send(message string, params *types.Params) error {
 	var firstErr error
 
+	splitLines := service.config.SplitLines
+	rawSplitLines, ok := (*params)["splitLines"]
+	if ok {
+		splitLines, _ = format.ParseBool(rawSplitLines, service.config.SplitLines)
+	}
+
 	if service.config.JSON {
 		postURL := CreateAPIURLFromConfig(service.config)
 		firstErr = doSend([]byte(message), postURL)
 	} else {
-		batches := CreateItemsFromPlain(message, service.config.SplitLines)
+		batches := CreateItemsFromPlain(message, splitLines)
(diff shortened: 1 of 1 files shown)


danielskowronski:
Quick and ugly test code:

```go
package main

import (
	"fmt"
	"os"

	"github.com/containrrr/shoutrrr"
)

func getUrls() []string {
	var urls []string
	urls = append(urls, os.Getenv("SHOUTRRR_URL"))

	return urls
}

func genMessage(params *types.Params) string {
	message := "This is line 1\nThis is line 2\nThis is line 3\n\n"
	for key, value := range *params {
		if key == "title" || key == "color" {
			continue
		}
		message += fmt.Sprintf("It was sent with key `%s` set to `%s`\n", key, value)
	}
	return message
}

func send(setting_name string, setting_value string) {
	title := "Test Title"
	urls := getUrls()
	sender, err := shoutrrr.CreateSender(urls...)
	if err != nil {
		fmt.Printf("create shoutrrr sender: %s", err)
		return
	}
	params := &types.Params{
		"title":      title,
		"color":      "0xFFFF00",
		setting_name: setting_value,
	}
	message := genMessage(params)
	errorsList := sender.Send(message, params)
	for _, err := range errorsList {
		if err != nil {
			fmt.Printf("send shoutrrr notification: %s", err)
			return
		}
	}
	fmt.Println("ok")
}

func main() {
	send("splitLines", "No")
	send("splitLines", "Yes")
}
```

